PR #16, opened 2021-09-29
Add EAV models and CRUD operations [WIP]

commanderprice:


backend/crud.py
```diff
@@ -0,0 +1,211 @@
...
+    SchemaUpdateSchema,
+    AttributeCreateSchema
+)
+
+
+def create_attribute(db: Session, data: AttributeCreateSchema) -> Attribute:
+    exists = db.execute(select(Attribute).where(Attribute.name == data.name)).scalar()
+    if exists:
+        # TODO custom exceptions
+        raise Exception(f'Attribute with name `{data.name}` already exists')
```
crazyscientist, line 28: As you suspected in the chat, this might be a candidate for _silent failure_.

I assume you had an alternative in mind like the `get_or_create()` method of Django QuerySets. That would indeed make sense to me.

Some additional thoughts:

* If this function gets exposed via the API, it would be nice to indicate the difference (attribute created or already existing) via different HTTP status codes (e.g. 201 and 200).
* We might hide this function even from clients. If a client requests the creation of an attribute definition (as part of a schema definition), we might already offer the name and type of the attribute instead of the attribute ID.

backend/crud.py
```diff
@@ -0,0 +1,211 @@
...
+    a = Attribute(name=data.name, type=data.type)
+    db.add(a)
+    db.commit()
+    return a
+
+
+def create_schema(db: Session, data: SchemaCreateSchema) -> Schema:
+    # TODO require lock here and in update?
```
crazyscientist, line 36: It's probably unlikely that there are simultaneous requests for creating schemas, but it's better to be safe than sorry. I would use a lock here :smiley:

backend/crud.py
```diff
@@ -0,0 +1,211 @@
...
+    exists = db.execute(select(Schema).where( (Schema.name == data.name) | (Schema.slug == data.slug) )).scalar()
+    if exists:
+        raise Exception(f'Schema with name `{data.name}` or slug `{data.slug}` already exists')
```
crazyscientist, line 39: Yes, I agree. We don't want ambiguous names or slugs.

As I'm more familiar with the Django ORM I'm wondering if it would be sufficient to rely on the unique constraints, which are defined in the `Schema` model, causing an exception during the attempt to save the data?

backend/crud.py
```diff
@@ -0,0 +1,211 @@
...
+    
+    sch = Schema(name=data.name, slug=data.slug)
+    db.add(sch)
+    db.flush()
+    for attr in data.attributes:
+        a: Attribute = db.execute(select(Attribute).where(Attribute.id == attr.attr_id)).scalar()
```
crazyscientist, line 45: If we used the `create_attribute()` function (with the `get_or_create_ equivalent mentioned in an earlier comment), we can indeed drop the following check and exception.

backend/crud.py
```diff
@@ -0,0 +1,211 @@
...
+            bfk = BoundFK(attr_def_id=ad.id, schema_id=s.id)
+            db.add(bfk)
+    db.commit()
+    return sch
+
+
+def delete_schema(db: Session, schema_id: int) -> Schema:
+    schema = db.execute(select(Schema).where(Schema.id == schema_id)).scalar()
+    if schema is None:
+        raise Exception('Should we raise exception here?')
```
crazyscientist, line 77: I would assume that a deletion of a schema will only be triggered by an API request. In this case, having a status code 404 seems to be a worthy goal. To achieve this, an exception seems like a good choice :+1:

crazyscientist: You might also want to check, whether the schema is already deleted. In that case a 404 response would also be warranted.

backend/crud.py
```diff
@@ -0,0 +1,211 @@
...
+    db.execute(
+        update(Entity).where(Entity.schema_id == schema_id).values(deleted=True)
+    )
+    schema.deleted = True # TODO require locks?
```
crazyscientist, line 81: It's probably overcautious, but I would use a lock.

backend/crud.py
crazyscientist: In a Django project I ran into problems with simultaneous calls to `get_or_create` methods. As a mere precaution I would suggest to use a lock here.

backend/crud.py
```diff
@@ -0,0 +1,211 @@
...
+
+        try:
+            if isinstance(value, list):
+                if not attr_def.list:
+                    raise Exception(f'Attribute {attr.name} on schema {sch.name} cannot hold multiple values')
+                values = [caster(i) for i in value]
+            else:
+                values = [caster(value)]
+        except ValueError as e:
+            raise # can this even happen with validation done before?
```
crazyscientist, line 187: I assume that you want to use a dynamically created pydantic model for validation. If my understanding is correct, this should also take care of the type casting. If that is the case, this explicit type casting and exception handling might become obsolete.

backend/crud.py
```diff
@@ -0,0 +1,211 @@
...
+            v = model(value=val, entity_id=e.id, attribute_id=attr.id)
+            db.add(v)
+    db.commit()
+    return e
+
+
+def delete_entity(db: Session, entity_id: int) -> Entity:
+    e = db.execute(select(Entity).where(Entity.id == entity_id)).scalar()
+    if e is None:
+        raise Exception('Should we raise exception here?')
```
crazyscientist, line 208: If requested via the API, the client should receive a 404, if the entity does not exist or is already deleted. So, I'd say an exception is a good choice :+1: 

And, yes, a lock makes sense to me.

crazyscientist (commented):
Thank you for sharing the current state

backend/models.py
```diff
@@ -0,0 +1,163 @@
...
+    __tablename__ = 'values_str'
+    value = Column(String)
+
+
+class ValueDatetime(Value):
+    __tablename__ = 'values_datetime'
+    value = Column(DateTime)
+
+
+class AttrType(enum.Enum):
```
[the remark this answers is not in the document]
commanderprice: It makes sense, I was also thinking about having everything in one place. Actually, I dropped in your definition of `AttrType` and none of my tests broke 😏. It will require to rewrite some stuff in crud.py, but I think we can go with approach like you suggested

backend/crud.py
commanderprice: > I assume you had an alternative in mind like the get_or_create() 

Actually I haven't thought about it😅 But yeah, it might be useful to go this way

> indicate the difference via different HTTP status codes

Makes sense

> would suggest to use a lock here

Makes sense

> We might hide this function even from clients

I don't think that this is possible — attributes have to appear from somewhere, right?

> If a client requests the creation of an attribute definition we might already offer the name and type of the attribute

That's actually how I was thinking it should work: firstly users define the attribute itself and when they create schema they can choose from list of existing attributes

backend/crud.py
commanderprice: My reason to put lock here (and in `update_schema()`) was that there might be simultaneous requests for creating and updating schemas which may lead to having multiple schemas with same name/slugs

backend/crud.py
commanderprice: > this should also take care of the type casting
Yeah, that's what was on my mind. Well, let's leave it they way it is until there is code for creating dynamic schemas😉

backend/crud.py
commanderprice: > If we used the create_attribute() function (with the `get_or_create_ equivalent mentioned in an earlier comment)

How this would stop me from providing data like
```
{
  ...
 "attributes": [
    {
       attr_id: 99999
       ...
    }
  ]
}
```
when there is no `Attribute` with provided id?

backend/crud.py
commanderprice: Well, we can go this way, but I'd rather prefer to raise custom exceptions and handle them correspondingly on API level

backend/crud.py
crazyscientist: Yes, that is one approach. When writing the comment, I was thinking that it would be nice for clients, if they would not need to deal with the attribute creation directly. I imaged a request with data like this for schema creation:

```json
{
  "name": "New Schema",
  "slug": "new-schema",
  "attributes": [
    {
      "name": "foo",
      "type": "str",
      "required": true
    },
    {
      "name": "bar",
      "type": "float",
      "list": true,
      "required": false
    }
  ]
}
```

So, during the schema creation, the backend would take care of creating the attributes as needed. So one the frontend the attributes can be defined freely.

What do you think? Can that work for you?

backend/crud.py
crazyscientist: Assuming that the SqlAlchemy ORM behaves like the Django ORM, maybe we can find a compromise:

We try saving the data to the database, catch the exception from SqlAlchemy and use the `raise ... from ...` syntax to get a custom exception on the API level for further handling?

backend/crud.py
commanderprice: Well, that will complicate things, but yes, that's possible

backend/crud.py
crazyscientist: Yes, that would require some changes. So the data would need to include the `name` and `type`, so that the attribute object can be created of obtained.

Could that work for you?

backend/crud.py
commanderprice: So, I checked that, we can wrap
```python
sch = Schema(name=data.name, slug=data.slug)
db.add(sch)
db.flush()
```
in try-except and catch `IntegrityError` from ORM and then raise our custom exception

backend/crud.py
commanderprice: Well, it should be implementable. I can't say whether it is a good idea to go this way, but we can try at least

crazyscientist:
Thank you for the latest updates.

Based on your usage of `pg_advisory_xact_lock` in the `crud` module, I assume you intend to handle the start, commit and rollback of transactions in a decorator, that will be applied to the functions from the `crud` module to make absolutely sure that other developers in the future cannot write to the DB outside of a transaction, which will not get rolled back in case of an error.

I like how you consider the different cases that need to be regarded for locking (e.g. prevent creation of an entity while a schema gets updated or deleted), but I spotted inconsistencies, e.g. the locks in the functions `update_schema` and `delete_schema` do not prevent the case that two different clients try to update and delete the same schema at the same time. Would you mind, if I try something?

commanderprice:
> Based on your usage of `pg_advisory_xact_lock` in the `crud` module, I assume...

Actually, I didn't plan to turn locking function into decorator and I am not sure if it makes sense to (maybe you have some use cases in mind?), but yes, I chose `xact_lock` because it is automatically released at the end of transaction or when session closes, so we don't need to think about releasing locks on our own

> to make absolutely sure that other developers in the future cannot write to the DB outside of a transaction, which will not get rolled back in case of an error

Maybe I got you wrong, but I don't think that locks are helping us with that, because even without them everything is supposed to work this way:

* Server gets request from client
* New session for this request is created
* This session is passed to crud functions which do their stuff and commit _once_
* If everything is ok, we do whatever we do, if not --  no data is supposed to be written, because all data changes are inside one transaction
* Session closes automatically regardless of whether an exception was thrown in crud

crazyscientist:
Maybe it's been too long since I last used SqlAlchemy.

With `autocommit=False` the transaction is not automatically commited, but rolled back. Yes, in this case we are safe. Sorry for the noise...

backend/exceptions.py
```diff
@@ -0,0 +1,94 @@
...
+
+class MissingAttributeException(MissingObjectException):
+    obj_type = 'Attribute'
+
+class MultipleAttributeOccurencesException(Exception):
+    def __init__(self, attr_name: str):
+        self.attr_name = attr_name
+    
+    def __str__(self) -> str:
+        return f'Found multiple occurences of attribute `{self.attr_name}`'
```
crazyscientist, line 37: Found a typo :wink: 

```suggestion
        return f'Found multiple occurrences of attribute `{self.attr_name}`'
```

backend/crud.py
```diff
@@ -0,0 +1,281 @@
...
+def update_schema(db: Session, schema_id: int, data: SchemaUpdateSchema) -> Schema:
+    sch: Schema = db.execute(
+        select(Schema).where(Schema.id == schema_id).where(Schema.deleted == False)
+    ).scalar()
+    if sch is None:
+        raise MissingSchemaException(obj_id=schema_id)
+    
+    schemas = db.execute(select(Schema).where( (Schema.name == data.name) | (Schema.slug == data.slug) )).scalars().all()
+    if len(schemas) == 1 and schemas[0].id != schema_id or len(schemas) > 1:
+        raise SchemaExistsException(name=data.name, slug=data.slug)
```
crazyscientist, line 125: We have defined the `name` and `slug` field as unique on the DB level. I'd suggest to rely SqlAlchemy to raise an exception and raise the custom exception from the `except:` block (as in `create_schema`).

crazyscientist (commented):
Looks good to me. I only found some candidates for minor enhancement :slightly_smiling_face:

backend/models.py
```diff
@@ -0,0 +1,159 @@
...
+    FK = Mapping(ValueForeignKey, int)
+    DT = Mapping(ValueDatetime, datetime.fromtimestamp)
+
+
+class BoundFK(Base):
+    __tablename__ = 'bound_foreign_keys'
+
+    id = Column(Integer, primary_key=True, index=True)
+    attr_def_id = Column(Integer, ForeignKey('attr_definitions.id'))
+    schema_id = Column(Integer, ForeignKey('schemas.id'))
```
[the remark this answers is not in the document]
commanderprice: Perhaps we should change `BoundFK.schema` name to `BoundFK.bound_schema` to make it more clear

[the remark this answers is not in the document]
crazyscientist: Oh, right. Sorry, I completely forgot about that. I don't think it's necessary to rename the field; it might be sufficient to add a `doc` or `comment` parameter to it.

commanderprice:
Fixed all stuff🔨

crazyscientist:
Great, let's merge it :grin:

PR #25, opened 2021-10-15
Add API routes

commanderprice:
Added some API routes to illustrate idea of how I see dynamic routes for schemas. The idea is that we have `create_dynamic_router()` (`dynamic_routes.py`) function that (given schema, FastAPI instance etc.) will create API routes for this schema. Routes are created at server start (`create_app()` in `__init__.py`) and when we create/update schema (currently `create_schema()` and `update_schema()` in `general_routes.py`). 

Routes logic is the same for all schemas, so main effect of `create_dynamic_router()` is to provide API documentation for these routes and dynamically generate input/output Pydantic models that are used for validation and API docs as well

### DIFF
--- a/backend/crud.py
+++ b/backend/crud.py
@@ -354,8 +354,10 @@ def _parse_filters(filters: dict, attrs: List[str]) -> Tuple[Dict[str, Dict[str,
         if attr == 'name':
             name_filters[filter] = v
             continue
-        elif attr not in attrs or filter not in FILTER_MAP:
-            continue
+        elif attr not in attrs:
+            raise InvalidFilterAttributeException(attr=attr, allowed_attrs=attrs)
+        elif filter not in FILTER_MAP:
+            raise InvalidFilterOperatorException(attr=attr, filter=filter)
         attrs_filters[attr][FILTER_MAP[filter]] = v
     
     name_filters = {k: v for k,v in name_filters.items() if k in FILTER_MAP and k in ALLOWED_FILTERS[AttrType.STR]}
@@ -367,7 +369,7 @@ def _query_entity_with_filters(filters: dict, schema: Schema, all: bool = False,
     to get entities that satisfy all conditions from `filters`
     '''
     
-    attrs = {i.attribute.name: i.attribute for i in schema.attr_defs}
+    attrs = {i.attribute.name: i.attribute for i in schema.attr_defs if not i.list and i.attribute.type in ALLOWED_FILTERS}
     attrs_filters, name_filters = _parse_filters(filters=filters, attrs=list(attrs))
     selects = []
 
--- a/backend/crud.py
+++ b/backend/crud.py
@@ -2,7 +2,7 @@ from typing import Callable, Dict, List, Tuple, Union
 from collections import defaultdict
 
 import sqlalchemy
-from sqlalchemy import func
+from sqlalchemy import func, distinct
 from sqlalchemy.orm import Session
 from sqlalchemy import select, update
 from sqlalchemy.sql.expression import delete, intersect
@@ -413,7 +413,7 @@ def get_entities(
         q = select(Entity).where(Entity.schema_id == schema.id)
         if not all:
             q = q.where(Entity.deleted == deleted_only)
-    total = db.execute(select(func.count('id')).select_from(q.subquery())).scalar()
+    total = db.execute(select(func.count(distinct(column('id')))).select_from(q.subquery())).scalar()
     q = q.offset(offset).limit(limit).order_by(Entity.id)
     entities = db.execute(select(Entity).from_statement(q)).scalars().all()
 
--- a/backend/crud.py
+++ b/backend/crud.py
@@ -66,12 +66,8 @@ def create_attribute(db: Session, data: AttributeCreateSchema, commit: bool = Tr
 
 def get_schemas(db: Session, all: bool = False, deleted_only: bool = False) -> List[Schema]:
     q = select(Schema)
-    if all:
-        pass
-    elif deleted_only:
-        q = q.where(Schema.deleted == True)
-    else:
-        q = q.where(Schema.deleted == False)
+    if not all:
+        q = q.where(Schema.deleted == deleted_only)
     return db.execute(q).scalars().all()
 
 
@@ -329,7 +325,7 @@ def _get_attr_values_batch(db: Session, entities: List[Entity], attrs_to_include
 
 FILTER_MAP = {
     'eq': '__eq__',
-    'lt':'__lt__',
+    'lt': '__lt__',
     'gt': '__gt__',
     'le': '__le__',
     'ge': '__ge__',
@@ -352,8 +348,8 @@ def _parse_filters(filters: dict, attrs: List[str]) -> Tuple[Dict[str, Dict[str,
     attrs_filters = defaultdict(dict)
     name_filters = {}
     for f, v in filters.items():
-        split = f.split('.')
-        attr = '.'.join(split[:-1]) if len(split) > 1 else split[0]
+        split = f.rsplit('.', maxsplit=1)
+        attr = split[0]
         filter = 'eq' if len(split) == 1 else split[-1]
         if attr == 'name':
             name_filters[filter] = v
--- a/backend/crud.py
+++ b/backend/crud.py
@@ -2,7 +2,7 @@ from typing import Callable, Dict, List, Tuple, Union
 from collections import defaultdict
 
 import sqlalchemy
-from sqlalchemy import func, distinct
+from sqlalchemy import func, distinct, column
 from sqlalchemy.orm import Session
 from sqlalchemy import select, update
 from sqlalchemy.sql.expression import delete, intersect
@@ -371,7 +371,7 @@ def _query_entity_with_filters(filters: dict, schema: Schema, all: bool = False,
     to get entities that satisfy all conditions from `filters`
     '''
     
-    attrs = {i.attribute.name: i.attribute for i in schema.attr_defs if not i.list and i.attribute.type in ALLOWED_FILTERS}
+    attrs = {i.attribute.name: i.attribute for i in schema.attr_defs if i.attribute.type in ALLOWED_FILTERS}
     attrs_filters, name_filters = _parse_filters(filters=filters, attrs=list(attrs))
     selects = []
 
--- a/backend/crud.py
+++ b/backend/crud.py
@@ -268,12 +268,8 @@ def get_entities(
     ) -> List[dict]:
     
     q = select(Entity).where(Entity.schema_id == schema.id)
-    if all:
-        pass
-    elif deleted_only:
-        q = q.where(Entity.deleted == True)
-    else:
-        q = q.where(Entity.deleted == False)
+    if not all:
+        q = q.where(Entity.deleted == deleted_only)
     entities = db.execute(q.offset(offset).limit(limit)).scalars().all()
 
     attr_defs = schema.attr_defs if all_fields else [i for i in schema.attr_defs if i.key]
--- a/backend/crud.py
+++ b/backend/crud.py
@@ -1,8 +1,8 @@
-from typing import Dict, List
+from typing import Dict, List, Union
 
 import sqlalchemy
 from sqlalchemy.orm import Session
-from sqlalchemy import select, update, func
+from sqlalchemy import select, update
 
 from .models import (
     AttrType,
@@ -24,7 +24,24 @@ from .schemas import (
 from .exceptions import *
 
 
+RESERVED_ATTR_NAMES = ['id', 'name', 'deleted']
+
+
+def get_attributes(db: Session) -> List[Attribute]:
+    return db.execute(select(Attribute)).scalars().all()
+
+
+def get_attribute(db: Session, attr_id: int) -> Attribute:
+    attr = db.execute(select(Attribute).where(Attribute.id == attr_id)).scalar()
+    if attr is None:
+        raise MissingAttributeException(obj_id=attr_id)
+    return attr
+
+
 def create_attribute(db: Session, data: AttributeCreateSchema, commit: bool = True) -> Attribute:
+    if data.name in RESERVED_ATTR_NAMES:
+        raise ReservedAttributeException(attr_name=data.name, reserved=RESERVED_ATTR_NAMES)
+
     attr = db.execute(
         select(Attribute)
         .where(Attribute.name == data.name)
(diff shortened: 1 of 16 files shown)


backend/general_routes.py
```diff
@@ -0,0 +1,76 @@
...
+    except exceptions.AttributeNotDefinedException:
+        pass
+    except exceptions.ListedToUnlistedException:
+        pass
+    except exceptions.MultipleAttributeOccurencesException:
+        pass
+    except exceptions.AttributeAlreadyDefinedException:
+        pass
+    except exceptions.NoSchemaToBindException:
+        pass
```
crazyscientist, line 74: I assume we will replace the `pass` statements with some exception raising?

backend/crud.py
```diff
@@ -39,6 +57,30 @@ def create_attribute(db: Session, data: AttributeCreateSchema, commit: bool = Tr
...
 
 
+def get_schemas(db: Session, all: bool = False, deleted_only: bool = False) -> List[Schema]:
+    q = select(Schema)
+    if all:
+        pass
+    elif deleted_only:
+        q = q.where(Schema.deleted == True)
+    else:
+        q = q.where(Schema.deleted == False)
```
crazyscientist, line 67: I believe you can simplify this block a little bit to something like this:

```python
if not all:
    q = q.where(Schema.deleted == deleted_only)
```

backend/models.py
```diff
@@ -104,14 +104,18 @@ class Schema(Base):
 
 class Entity(Base):
     __tablename__ = 'entities'
-
+    
     id = Column(Integer, primary_key=True, index=True)
-    name = Column(String(128), nullable=False)
+    slug = Column(String(128), nullable=False)
```
crazyscientist, line 109: Actually, I think it makes sense to have both: `name` and `slug` :thinking:

commanderprice:
>The dynamic models for entities are not shown in the docs

If you are talking about schema description with fields and stuff then yes, it's not there yet, but it will be

> Probably I got something wrong when I tried to create an entity

Can't really say anything regarding this, but there will be tests to ensure that everything working properly 👌

backend/models.py
commanderprice: What `name` will be used for in this case? I mean, we can use `slug` both as `id` replacement and as something like title for entity to display it on frontend

commanderprice: > validator for slug to make sure it's indeed URL-friendly

This one is actually needed

> `name` is the human-friendly identifier/name/title.

Well, on frontend we can transform slug and replace `-` and `_` with spaces for example😏

crazyscientist: Yes, that is an option.

Just thinking out loud here: For users it might be more intuitive, if it's the other way. I'm thinking of this workflow:

* User fills out an input form (to create an entity).
* While typing in the field for `name`, the field for `slug` gets automatically filled.
* When editing an entity, users can change `name` and `slug` independently.

This way we would set some decent defaults and the entity URL (with slug) would not change, if the name actually needs to change.

In addition, we could make  `(schema_id, slug)` unique and allow names to be non-unique :thinking:

commanderprice: Alright, but now we get back to one of my previous questions

> allow names to be non-unique

Since `name` works as, basically, predefined attribute, users should be able to make it either unique or non-unique on specific schema. Should we add then a flag to `Schema` that would show uniqueness of `name`?

crazyscientist (commented):
Sorry, I did not manage to look at everything yet...

backend/crud.py
commanderprice: Oh, sorry, I applied your change in another function which had the same lines 😅 I'll change this one too

backend/crud.py
```diff
@@ -197,14 +265,231 @@ def update_schema(db: Session, schema_id: int, data: SchemaUpdateSchema) -> Sche
...
+
+
+def _parse_filters(filters: dict, attrs: List[str]) -> Tuple[Dict[str, Dict[str, Any]], Dict[str, Dict[str, Any]]]:
+    '''Returns tuple of two `dict`s like `{attr_name: {op1: value, op2: value}}`.
+    First `dict` is for attribute filters, second is for `Entity.name` filters
+    '''
+    attrs_filters = defaultdict(dict)
+    name_filters = {}
+    for f, v in filters.items():
+        split = f.split('.')
```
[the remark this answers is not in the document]
commanderprice: Oh, that's why it felt wrong to me when I was reinventing the wheel 😅 Yes, `rsplit` is definitely better here

backend/crud.py
```diff
@@ -197,14 +265,231 @@ def update_schema(db: Session, schema_id: int, data: SchemaUpdateSchema) -> Sche
...
+    'eq': '__eq__',
+    'lt':'__lt__',
+    'gt': '__gt__',
+    'le': '__le__',
+    'ge': '__ge__',
+    'ne': '__ne__'
+}
+
+ALLOWED_FILTERS = {
+    AttrType.STR: ['eq', 'lt', 'gt', 'le', 'ge', 'ne'],
```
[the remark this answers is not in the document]
commanderprice: 1. For matching substring (and not matching it) we have two options: `like` and `contains`. We can simulate `contains` with `like` by writing queries like `%substr%` and we have an opposite method, `not_like`, to search for strings that don't satisfy condition. On the other hand, `contains` doesn't have `not_like` alternative
2. There is a `regexp_match` method for... matching regexps and, as in case with `contains`, it doesn't have a negation variant, which means that filter inside `where` has to be negated with operator `~`. To add such negation we have to make some changes to code.

Option 1. Create `NEGATED_FILTERS` list so that when in `_query_entity_with_filters` we are iterating over filters we could check if filter is inside this list, and if it is then negate `where` condition

Option 2 sounded better inside my head, so we are left only with Option 1, but maybe you have some ideas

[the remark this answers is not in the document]
crazyscientist: 1. We can expose filters `contains` and `notcontains` on the API, but use the `like` and `not_like` functions as you suggested.
2. Having the `regexp_match` might already be a good start.

> but maybe you have some ideas

I had the exact same problem with Django: It supported filtering rows by matching a RegExp, but not the opposite (filtering by not-matching). It was not too difficult to add a wrapper function in Python to make this feature available. I haven't used SqlAlchemy in a long time, but I could imagine that something similar is possible. But we don't need to worry about it now...

backend/crud.py
```diff
@@ -197,14 +265,231 @@ def update_schema(db: Session, schema_id: int, data: SchemaUpdateSchema) -> Sche
...
+    name_filters = {}
+    for f, v in filters.items():
+        split = f.split('.')
+        attr = '.'.join(split[:-1]) if len(split) > 1 else split[0]
+        filter = 'eq' if len(split) == 1 else split[-1]
+        if attr == 'name':
+            name_filters[filter] = v
+            continue
+        elif attr not in attrs or filter not in FILTER_MAP:
+            continue
```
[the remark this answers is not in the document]
commanderprice: It depends on whether you see passing filter with wrong attr. name/operator as a huge mistake that requires raising an exception

backend/crud.py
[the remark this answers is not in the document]
commanderprice: > but use the like and not_like functions as you suggested.

Do you mean that when user sends `attr.contains=substr` backend should use `attr.like('%substr%')`? If so, this is not going to work (at least in current implementation) because that would require to modify query value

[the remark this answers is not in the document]
crazyscientist: Yes, that is what I had in my mind :joy: 

I'd love to see this in the MVP, but we don't need to implement it right away. We can postpone it for a while.

backend/crud.py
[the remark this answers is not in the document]
crazyscientist: I agree, it will not be a _huge mistake_. I'm only worried that it would be very difficult to debug client code, if the attribute or operator name is incorrect. So I would have been in favor of presenting an error 4xx to the client, which complains about the invalid/unknown filter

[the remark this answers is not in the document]
commanderprice: Alright, I'll add an exception then

backend/schemas.py
```diff
@@ -28,29 +29,107 @@ class AttributeDefinitionBase(BaseModel):
...
-    id: int
+    attr_def_id: int
+
+
+class AttributeDefinitionUpdateWithNameSchema(AttributeDefinitionBase):
+    name: str
+
+
+def validate_slug(cls, slug: str):
+    if re.match('^[a-zA-Z0-9]+(?:-[a-zA-Z0-9]+)*$', slug) is None:
```
crazyscientist, line 60: I don't think you need the `?:`. A capturing group should work just as well.

After looking at the handling of `id_or_slug` in the CRUD functions, I'm afraid we need to address one issue with this pattern:

The pattern allows numeric slugs like e.g. `123`. If my understanding is correct, the CRUD functions would treat this as ID instead of a slug. We might need a slightly more complicated pattern, e.g.:

```suggestion
    if re.match('^([a-zA-Z][-a-zA-Z0-9]+)|([0-9]+[-a-zA-Z]+)$', slug) is None:
```

commanderprice: > A capturing group should work just as well

Yeah, it seems to be so

> would treat this as ID instead of a slug

Right, haven't thought about it. I'll change regexp to `^[a-zA-Z]+(-[a-zA-Z0-9]+)*$`, you can take a look at it here https://regex101.com/r/92FRVJ/2

crazyscientist: Looking at your test strings, I would object strongly only to `456` and those starting with `-`; maybe even `hello-$-world`.

If you want to exclude slugs starting with a number, it's also fine.

commanderprice: > object strongly only to...

Well, I don't want to allow slugs that have multiple `-` in a row or which end with `-` because they just look wrong to me

> If you want to exclude slugs starting with a number

Yes, that was my idea

backend/tests/test_dynamic_routes.py
```diff
@@ -0,0 +1,600 @@
...
+            'born': '2021-10-20T13:52:17',
+        }
+        response = client.post(f'/person', json=p2)
+        assert response.json() == {'id': 4, 'slug': 'John', 'name': 'john', 'deleted': False}
+
+        john = dbsession.execute(select(Entity).where(Entity.id == 4)).scalar()
+        assert john.get('nickname', dbsession).value == 'john'
+        assert john.get('age', dbsession).value == 10
+        assert [i.value for i in john.get('friends', dbsession)] == [3, 1]
+        assert john.get('born', dbsession).value == datetime(2021, 10, 20, 13, 52, 17)
```
crazyscientist, line 265: I had not noticed before, or I was just assuming, but timestamps without timezone might become confusing.

If it's not too much effort, I'd appreciate a change to use timezones (and set a default timezone, if none is provided). But we could address this in a separate issue/PR.

backend/schemas.py
crazyscientist: That will work for me :+1:

backend/tests/test_dynamic_routes.py
commanderprice: Ok, so, I changed `value = Column(DateTime)` to `value = Column(DateTime(timezone=True))` in `ValueDatetime` and now PG saves date in column of type `timestamp with time zone` and converts all datetimes to UTC. Will that work for you?

crazyscientist: That will work... If we want another default timezone (i.e. one that differs from the default timezone of the DB) in the future, we could address that in the field validation, right?

commanderprice: > If we want another default timezone

When doing what? Getting/passing dates from/to backend? I didn't really get the question🤔

commanderprice: If your question is "Is it possible to return dt in timezone other than defined in DB by using pydantic validators?" then yes, I guess so

crazyscientist: I was just thinking out loud: If I understood you correctly, if a timestamp does not define the timezone, the database uses it's default timezone. In a worst case scenario, where we cannot change the timezone of the database, I was thinking about adding a default timezone as part of the value validation.

But I think, the timezone configured in the database should be sufficient.

crazyscientist: > If your question is

Yes, that was the idea. But I don't think we need to implement this now.

backend/crud.py
```diff
@@ -197,14 +261,235 @@ def update_schema(db: Session, schema_id: int, data: SchemaUpdateSchema) -> Sche
...
+    name_filters = {k: v for k,v in name_filters.items() if k in FILTER_MAP and k in ALLOWED_FILTERS[AttrType.STR]}
+    return attrs_filters, name_filters
+
+
+def _query_entity_with_filters(filters: dict, schema: Schema, all: bool = False, deleted_only: bool = False) -> CompoundSelect:
+    '''Returns intersection query of several queries with filters
+    to get entities that satisfy all conditions from `filters`
+    '''
+    
+    attrs = {i.attribute.name: i.attribute for i in schema.attr_defs if not i.list and i.attribute.type in ALLOWED_FILTERS}
```
crazyscientist, line 374: If I'm reading this correctly, we are excluding all list-type attributes from filtering. I don't think we should do that :thinking:

crazyscientist: It seems I was already overlooking this limitation in `dynamic_routes._filters_request_model`.

commanderprice: Yes, I prohibited filtering on listed fields because... I don't actually remember why🙄 If you think that we should be able to filter on lists as well (when we check all values for provided condition) then I'll take a look later to see if it's going to work well if I just remove that `if`

crazyscientist: > If you think that we should be able to filter on lists 

Yes, I think this would make sense :grin: 

> when we check all values for provided condition

This might be counter-intuitive for users. I believe it would be more intuitive, if we just used the existing query for lists as well. In other words, on lists it's sufficient that any value in the list fulfills the filter condition(s).

> see if it's going to work well if I just remove that `if`

Thank you. If it does not go that well, we can address this in a separate PR.

commanderprice: > it's sufficient that any value in the list fulfills the filter condition

Yeah, that was my point as well, I just meant that we'll have to check many values from list until we (don't) find satisfying our condition. And that also leads to question: what filters do we use in listed attrs? Just the ones which apply to underlying data type? i.e. if we have list of strings then for this attribute we can use all filters which we can use for string attributes

commanderprice: So, I've allowed filtering lists and that's how it works now. Consider we have Jack and Jane with attr `fav_color` equal to `['red', 'blue']` and `['red', 'black']` correspondingly. We'll get such results for following queries:
* `fav_color.contains=b` --> Jack, Jane (`blue` and `black`)
* `fav_color.eq=black` --> Jane
* `fav_color.ne=black` --> Jack, Jane (Jane has `black`, but also has `red` which satisfies our condition)

I don't like `not equal` works in this case, but I guess it the most we can get without writing logic specifically for lists

crazyscientist: > I don't like `not equal` works in this case

Yes, I can see why :joy: I'd say we keep it like that and we address this via a separate issue and PR.

crazyscientist: Created #27.

crazyscientist (approved):
Thank you. This PR looks good to me :+1:

PR #29, opened 2021-11-30
Use `alembic` for database migrations

crazyscientist:
* Updated `README`
* Moved definition of `SQLALCHEMY_DATABASE_URL` from `database` to `config`
* Requirements
  * Added `alembic` and `uvicorn`
  * Removed frozen version strings


commanderprice (approved):
Okay, I can't say anything specific about config of Alembic, but I've run and it looks like it can generate and apply migrations👍

crazyscientist:
> But probably you should remove existing migration in your files before merging

Well, that is the point of defining migrations. The migration in this PR represents the current data model. With future changes to the data model we will need to create more migrations (thankfully this is automated :joy:).

commanderprice:
So, your idea is to create new migrations when there is a change in data model and to commit both? Okay, haven't thought about it

crazyscientist:
> So, your idea is to create new migrations when there is a change in data model and to commit both? 

Exactly, I'm so used to Django, where this is the default behavior. Sorry, it did not occur to me that you might not be familiar with such a practice.

commanderprice:
I was just thinking about using migrations as way to apply model changes to database😅

crazyscientist:
I'm not 100% sure how alembic handles it, but Django needs the old migration files to calculate the diff for new migrations. But it feels logical that alembic would to it the same way.

And thank you for approving 🙂

PR #31, opened 2021-12-01
Add Traceability, Auth & Frontend

commanderprice:
What's new:

# Models

* added value table for `DATE` type;
* created migration for current state of models. Alembic seems to suck at enums, so in migration's `upgrade()` I had to manually update `attrtype` enum. Also it added a couple of new enums to database, which won't be removed by `downgrade()`  ¯\\_(ツ)\_/¯ (unless manual drop statements are included). And it won't remove `DATE` from `attrtype` because you can't drop values from enums.
* added user model;
* added traceability models;

New traceability models differ a little bit from what we discussed earlier, [here](http://www.plantuml.com/plantuml/uml/jLD1RXmX4BppYlw0dv38aPEp3uvpYQmHixHi4c3MGsDhPUtVGqCMc116IQJSXYgwfgkwkVOaLmhM22kFSFCatIE-2G2juDrxq8xYzo6XQffFZDrxN2Euuel6ruBU7aalJfglJ3zmqOgICAce-47QOgdgO4-IWcSmTt2NpkNNqp9_mmDL_ZQTAq-d8rRsSpmnjrYBZZRTUOhDYRD8RkyEJsZbJMhqhOXLdGWvQHCGwuGtuV7odNt8sIHsNXRpY-hzKOrKVRh0GBuQxon6JIpVz-hmTS3agjnWpyMw0STCstx3iuy7mRBrGi-pW4A3XEgtnKEI8yePVI1QTn82aF420tnHpu7oxZm7lQ8gYGCOxQav1gUV0pR0Tpmr9uN-iEhKKX_Bh9sQQ9dAT2gl7Ur7Ugwk_dYR02vBosMyil_T8i2-oOsHZsYGZVG--k6KkNwH9k3EOD6HfjJlf_Gbv4LYg_SpxHUMgu7nRhNGfTsizHV9joyPe7kF6zdYfbVjt0mt17RUyV0L0rE3edz_WpmgKGSCLrV5KDdV3iqXj6YEQeJtAircjB0GERg-8EDznJKw5Qpv0G00) you can see diagram for current state. The main changes are that both update and delete of entity/schema are handled by `XXXUpdate` model and creation and updates of values are handled just by `ValueUpdate` model. Also, in `Change` we now store object of change (e.g. `SCHEMA`) and type of change (e.g `UPDATE`). By the moment I'm writing this I don't really remember why I added that, but let's pretend that it's for making traceability more foolproof.

# Traceability

* added functions to create/apply/decline create/update/delete requests for schemas/entities. 
* added functions to get N recent changes for object (probably needs `offset` option as well)(return only changes in objects, not creation events) 
* added functions to get details for change (don't handle creation events)
* added corresponding routes for these functions
* CRUD routes were updated to use these functions
  * They require a user to be passed, but since auth is not there yet, these routes now have a placeholder code which gets and passes admin user
* added tests for this stuff, but I think still not for everything

# API
* dynamic routes now have prefix `/dynamic` and no longer require to care about reserved schema slugs;
* routes for entity/ies retrieveal now can optionally pass some metadata needed by frontend (and now I probably need to add another reserved name for attributes 🤔)
* previously create/update routes provided ability to pass data either by attr/attr_def id or by attribute name. I left only the second option for less confusion and branching in code
* entities now can be ordered by `name` and by user-defined fields. If field is non-key, it will still be returned in results;
* schema updates now can handle renaming and deletion of attributes;

# Frontend
* all pages currently have neither validation of user data nor handling of errors received from server;
* no tests;
* added entity list page for switching between schemas and searching for entities;
* requires `VUE_APP_API_BASE` env variable with value of API base URL w/o trailing slash. Env vars can be defined in `.env` files in `frontend` folder;
* added entity detail page;
* added entity create and update pages;
* added entity changes details page;
* added schema detail page;
* added schema create page;
* added schema update page;
* more on its way

# Auth, Frontend & refactoring
are on their way


backend/tests/test_crud_schema.py
```diff
@@ -364,288 +313,280 @@ def test_update(self, dbsession):
...
+             AttrDefUpdateSchema(
+                    name='nickname',
+                    new_name='fav_color',
+                    required=False,
+                    unique=False,
+                    list=False,
+                    key=False,
+                    description='updated'
+                )
+        ]
```
crazyscientist, line 432: The purpose of this statement eludes me. Did you intended to test more?

backend/tests/test_crud_schema.py
```diff
@@ -691,14 +629,17 @@ def test_raise_on_multiple_attrs_with_same_name(self, dbsession):
...
                     required=True,
                     unique=True,
                     list=True,
                     key=True,
                     bind_to_schema=-1
                 ),
                 AttrDefSchema(
+                    name='address',
+                    type='FK',
                     attr_id=address.id,
```
crazyscientist, line 643: In the above case, you have replaced `attr_id` with `name` and `type`. Here you have added them. I suspect you wanted to replace it here, too?

```suggestion
                    name='address',
                    type='FK',
```

crazyscientist (commented):
I got through half the test modules.... Will continue tomorrow :slightly_smiling_face:

backend/tests/test_crud_schema.py
commanderprice: Oh, actually that was for testing with updated data, but when I updated API, this test was no longer necessary, but I forgot to remove data for it😅

backend/tests/test_crud_entity.py
```diff
@@ -407,33 +438,13 @@ def test_update(self, dbsession):
             'friends': [1, 2],
         }
         update_entity(dbsession, id_or_slug=1, schema_id=1, data=data)
```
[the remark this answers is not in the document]
commanderprice: I expected that this one may mislead someone😅 `asserts_after_entities_update` contains checks that need to be run after _both_ updates

backend/tests/test_crud_schema.py
```diff
@@ -63,98 +89,10 @@ def data_for_test(self, db: Session) -> dict:
         }
 
     def test_create(self, dbsession):
-        data = self.data_for_test(dbsession)
+        data = TestSchemaCreate.data_for_test()
```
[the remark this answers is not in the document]
commanderprice: I guess that's just how I tend to call static methods🤔 Other reason may be that at that time I was factoring out common asserts into static methods (which later got factored out into functions out of classes) and in other test classes I had to call those methods by writing class name, so maybe I wrote that by inertia

backend/models.py
```diff
@@ -171,11 +178,236 @@ class AttributeDefinition(Base):
...
+    old_name = Column(String(128), nullable=True)
+    new_slug = Column(String(128), nullable=True)
+    old_slug = Column(String(128), nullable=True)
+    new_reviewable = Column(Boolean, nullable=True)
+    old_reviewable = Column(Boolean, nullable=True)
+    new_deleted = Column(Boolean, nullable=True)
+    old_deleted = Column(Boolean, nullable=True)
+
+    change = relationship('Change')
+    schema = relationship('Schema')
```
crazyscientist, line 328: Defining the `Change` table and separate tables for every tracked model and action seems robust but a bit difficult to maintain (in case somebody adds/removes columns to/from the models).

Maybe we could use a similar structure inspired by EAV? What would you think of something like this:

diagram

backend/traceability/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+
+def decline_change_request(db: Session, change_id: int, change_object: ChangeObject, change_type: ChangeType, reviewed_by: User, comment: Optional[str]) -> Change:
+    change = db.execute(
+        select(Change)
+        .where(Change.id == change_id)
+        .where(Change.change_object == change_object)
+        .where(Change.change_type == change_type)
+    ).scalar()
+    if change is None:
+        raise Exception('NO CHANGE NO CHANGE')
```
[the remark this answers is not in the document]
commanderprice: Yep, I currently have a couple of placeholder exceptions to be fixed a bit later

backend/models.py
commanderprice: 1. How can this structure be used to track changes in schemas?
2. How can we use it to track creation/deletion of schemas/entities?

It looks like after answering these questions a couple of other tables will be added, leading to very similar model that is there now 🤔 Also, when designing my models I thought that other objects can be tracked (changes in groups, users etc.) and if it so (at least potentially)

crazyscientist: > Also, when designing my models I thought that other objects can be tracked (changes in groups, users etc.)

Yes, that sounds like a good idea. In the long run we will probably connect to an authentication provider like LDAP or OpenID anyway. I'm not sure how much benefit such a tracking would provide.

However, we actually might be able to track users or other models with my proposal, too. :thinking: 

>  leading to very similar model that is there now

Yes, that could be considered a downside. On the plus side we would not need to worry about conversions between data types; we would also store values as their intended data type.

> How can we use it to track creation/deletion of schemas/entities?

In the `Change` model the type of the change (create, update, delete) would be indicated by the `type` field (or `change_type` if you prefer this name). The model would be specified in the `content_type` field. The `obejct_id` column would store the ID for all actions, except for creations.

Also it would store the column name **or** attribute ID that gets changed. We would also need to store the data type again (not only in `AttributeDefinition`; just in case somebody removes an attribute and adds an attribute with identical name but different data type later).

The `ChangeValueXXX` tables would store the actual old and new values (blank `new_value` for deletions, blank `old_value` for creations).

> How can this structure be used to track changes in schemas?

Indeed, this becomes a little more difficult as schema changes might include changes to attribute definitions. By moving the columns `content_type` and `object_id` to the `Change` model, we could track arbitrary changes to any model.

Let's assume the slug and an attribute definition of a schema change, we would create one `ChangeRequest` instance and

* a `Change` instance, where `content_type` is the Enum for schemas, `object_id` is the schema ID, `fieldname` is "slug" and the related `ChangeValueStr` holds the old/new value.
* a `Change instance`, where `content_type` is the Enum for attribute definitions, `object_id` is the ID of the attribute definition, `fieldname` is the name of the column/field to be changed and the related `ChangeValueXXX` holds the old/new value.

One thing I have not figured out, yet: If an attribute gets removed from a schema, how to keep track of it in the change log of the schema? Maybe it could be as simple as marking attribute definitions as deleted (instead of actually dropping from the table), too? :thinking: 

The updated data model could look like this:

diagram

What do you think?

commanderprice: Before I answer this question, can you tell me, how to handle attribute renaming in this case?

crazyscientist: >  how to handle attribute renaming in this case?

Renaming is basically just a replacement of the attribute IDs. To track the change in the schema, the `Change` instance would store the value "attribute_id" in the `fieldname` column. In `ChangeValueInt` we store the old and new ID of the attribute.

I admit, this might look a bit confusion as we also have the `attribute_id` field in the `Change` table :/

commanderprice: Do we use `attribute_id` in `Change` only for change of value for user-defined entity fields?

crazyscientist: That was the idea, yes. But we probably could use `fieldname` for attributes, too :thinking:

commanderprice: Okay, so, I guess I wrapped my mind (more or less) around your approach. I guess it can work, probably it's more flexible than current approach🤔 But keep in mind that for it I'll have to rewrite literally all the backend code related to traceability 🤪 I'm okay with that (although it's kind of frustrating), but that will take some time

crazyscientist: I'm not insisting on changing the approach. Primarily I wanted to share my current impression (not having looked at all the changes in the PR).

If you think it's worth the effort and you would like to change the approach, fine with me. If you prefer to stay on the present course, I'll simply save the idea in an issue in case we ever need to touch this particular part.

commanderprice: My reasons to keep current approach (apart from having to rewrite quite an amount of code) is that it's feels...simpler? On the other hand, I wasn't really thinking about someone deleting columns from models (handling adding new columns feels more straightforward), and your approach handles that by not referencing any particular column from any table, which also allows to connect new tables to traceability just by adding new values to some enums. So, if you feel that there will be changes in existing models and/or it will be useful to have such flexibility, that let's (try to) go with this option

crazyscientist: > that it's feels...simpler? 

Yes, I would have called it "straight forward" and "concrete".

>  So, if you feel that there will be changes in existing models and/or it will be useful to have such flexibility,

I'm not aware of upcoming changes, but who knows what new features, which will require changes to the data model, will be requested in the future :innocent: 

If we do not need to worry about the data model of the tracking tables, it would certainly be helpful. Assuming we ever have to touch one of the tracked tables.

But if you made good progress with the tracking feature, changing direction might not be worth the effort. As I'm not working actively as a developer on the project I can't judge, whether it's worth to change direction now. As primary developer, feel free to make the call. :grin:

commanderprice: > whether it's worth to change direction now

Well, currently traceability exists "on its own", meaning that only traceability (for the most part) depends on anything related with it, so these changes will hurt only traceability itself. I don't know if traceability will become "closer" to other components of the application in future, but if this happens and there will be a need to change data model, then it will be a more painful experience than if done now🤔

crazyscientist: My best guess: the two parts won't get closer by design; most likely future changes might require more model columns being tracked :thinking:

commanderprice: Okay, then I guess we should try to go with this approach. I think I won't start reimplementing traceability at least before next week, so maybe you'll have more ideas in the meantime😉

backend/crud.py
```diff
@@ -404,33 +443,68 @@ def get_entities(
         all: bool = False, 
         deleted_only: bool = False, 
         all_fields: bool = False,
-        filters: dict = None
+        filters: dict = None,
+        order_by: str = 'name',
+        ascending: bool = True,
     ) -> EntityListSchema:
-    
+    if order_by != 'name':
```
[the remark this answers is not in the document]
commanderprice: Not really, I didn't mean to allow users order by slug🤔

backend/crud.py
```diff
@@ -440,9 +514,9 @@ def _convert_values(attr_def: AttributeDefinition, value: Any, caster: Callable)
     if isinstance(value, list):
         if not attr_def.list:
             raise NotListedAttributeException(attr_name=attr_def.attribute.name, schema_id=attr_def.schema_id)
-        return [caster(i) for i in value]
+        return [caster(i) for i in [j for j in value if j is not None]]
```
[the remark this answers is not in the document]
commanderprice: Jesus, why did I write it that way😅

backend/crud.py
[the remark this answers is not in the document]
crazyscientist: OK, ordering by slug probably offers little value anyway. If somebody requests it, it should be easy to implement in the future.

backend/models.py
commanderprice: Oh, now I'm thinking where to store schema id when creating request for entity creation. Probably it can be placed in `Change.object_id`, but it looks like a misuse of field🤔

crazyscientist: Sorry for my late reply. But I would also choose the second option :

crazyscientist:
I just noticed a little inconsistency: 

In the schema API response the attribute types are defined in upper case, e.g.:

```json
# /schemas/foo-bar
{
  "id": 5,
  "name": "Foo Bar",
  "slug": "foo-bar",
  "deleted": false,
  "reviewable": false,
  "attributes": [
    {
      "name": "blabla",
      "type": "INT",
      "required": false,
      "unique": false,
      "list": false,
      "key": true,
      "description": "jhgjhgkg",
      "bind_to_schema": null
    },
    {
      "name": "hello",
      "type": "STR",
      "required": false,
      "unique": false,
      "list": false,
      "key": false,
      "description": "Say something",
      "bind_to_schema": null
    }
  ]
}
```

But in the `meta` section of entity listing the type is defined in lower case, e.g.:

```json
# /dynamic/foo-bar?meta=true
{
  "total": 3,
  "entities": [...],
  "meta": {
    "filter_fields": {
      "operators": {
        "int": [...],
        "str": [...]
      },
      "fields": {
        "blabla": {
          "type": "int"
        },
        "printu": {
          "type": "str"
        }
      }
    }
  }
}

**Comment by commanderprice** [comment]
Oh, it's not even about casing: first types are "instances" of our `AttrType` and these lowercase ones are names of Python datatypes derived from `AttrType.value.model.value`🤯. Currently I can't say why I would need a Python datatype here, so probably we'll just change it to `AttrType` names instead

File: backend/models.py (hunk 5 above)
Comment (commanderprice) [reply backend/models.py:333]: Now I have another issue with creating change request for schema creation. So, among other things, we need to store data of AttributeDefinitions of Schema, which include type, name, flags (required, list, etc.) etc. So, `Change` object for each attr.def. can look like this:
```python
Change(
      change_request=change_request,
      field_name='type',
      value_id=v.id, 
      data_type=ChangeAttrType.STR,
      content_type=ContentType.ATTRIBUTE_DEFINITION,
      change_type=ChangeType.CREATE
)
```
We can easily get `Change` objects for attr.def. creation associated with specific change request, but how to differentiate between different attributes?🤔 The only idea that came to my mind with current data model is to include `object_id=Attribute.id` for each `Change` of each AttributeDefinition, but that would require to create new `Attribute`s during request creation of they don't exist at the moment. What do you think?

crazyscientist: We had defined that it is very important to never update or delete attributes. The only _risky_ part is that clients create an obscene amount of attributes.

Long story short: For me it would be OK to create `Attribute` instances during request creation.

I could also think of the option to store the attribute name here, but that might complicate things too much :thinking:

commanderprice: Ok, so, I am in the middle of the process, but decided to commit what is already there so that you can take a look.
You can look at
* New models for traceability (at the bottom part of `models.py`, line 274)
* All updated functions for entities in `traceability/entity.py`
* `create_schema_create_request` and `apply_schema_create_request` in  `traceability/schema.py`

crazyscientist: Models look good to me :+1: 

I had a quick look at the functions for entities. I think this is the behavior I had in mind; but I noticed some duplicate or very similar lines, maybe we can reduce code duplication at some time in the future :wink: 

And I noticed that you already avoided this kind of code duplication in `create_schema_create_request` :+1: 

Yes, I think we are on the right track :upside_down_face:

commanderprice: Now that's kind of all I wanted to change in traceability (if I didn't forget about anything). A few things I wanted to mention:

* In models we have `ChangeObject` model which was used by initial traceability code extensively. Now it's used just by route for reviewing changes. It can be replaced by `ContentType` model because one is kind of a subset of another, but I'm not sure about it
* In `Change` model we have `value_id` as non-nullable. However there are cases when we don't really need to have value, for example when we are deleting attribute from schema (`traceability/schema.py`, line 337). Maybe it makes sense to make it nullable, but probably the case I described above won't happen too often

backend/models.py
```diff
@@ -1,15 +1,15 @@
 import enum
-from datetime import datetime
 from typing import List, Union, Optional, NamedTuple
+from datetime import datetime
 
 from sqlalchemy import (
-    select, Enum, DateTime,
+    select, Enum, DateTime, Date,
     Boolean, Column, ForeignKey, 
     Integer, String, Float)
```
crazyscientist, line 8: I'm not sure why, but these imports did not work well on my machine. Maybe you would like to rewrite them to:

```python
from sqlalchemy import select, Column, ForeignKey
from sqlalchemy.types import Enum, DateTime, Date, Boolean, Integer, String, Float
```

Also, you probably would like to create a migration for these changes :wink:

commanderprice: > did not work well on my machine

What was exactly the problem?

crazyscientist: I created a migration with alembic and tried to apply it. That failed because the it (the DB or sqlalchemy) claimed that it did not know the datatype `contenttype`.
After changing the imports like that and re-creating the migration, it was successfully applied. I admit, it's a bit weird :thinking:

commanderprice: Did you generate a new migration on top of the second (the one I commited for previous version of traceability)? And did you use command `alembic revision --autogenerate`? For some reason I get migrations with empty methods🤔

crazyscientist: Yes, that was my approach and the command actually created a non-empty migration.

Now that I think about it, it would probably make sense to replace the existing migration :thinking: 

Did you run the command in the  repo's root directory? That might also impact the behavior of alembic

commanderprice: > Did you run the command in the repo's root directory

I did

So, I finally had to use radical measures which was to drop all tables, move models to state before this PR, then upgrade to second migration, then insert new models and generate (now successfully) new migration, add some statements there to create enums (it's almost 2022 soon, why alembic can't create enums on its own?🤔) and finally apply this migration.

I've committed it (along with changes in models imports). I can't say for sure that migration worked correctly by running tests, because they use other database which is recreated all the time. But at a quick glance all tables are present and frontend seems to be working well with backend

crazyscientist:
I just stumbled across another problem: I have a reviewable schema called `foo2`. I have added the attribute `new` to the schema and created a change request for entity `hello-world` including this attribute. Afterwards I removed the attribute again. When I try to look at the change details I get this traceback:
```
INFO:     127.0.0.1:42684 - "GET /changes/entity/foo2/hello-world/40 HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/uvicorn/protocols/http/h11_impl.py", line 373, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/uvicorn/middleware/proxy_headers.py", line 75, in __call__
    return await self.app(scope, receive, send)
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/fastapi/applications.py", line 208, in __call__
    await super().__call__(scope, receive, send)
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/starlette/applications.py", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/starlette/middleware/cors.py", line 92, in __call__
    await self.simple_response(scope, receive, send, request_headers=headers)
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/starlette/middleware/cors.py", line 147, in simple_response
    await self.app(scope, receive, send)
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/starlette/exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/starlette/routing.py", line 656, in __call__
    await route.handle(scope, receive, send)
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/starlette/routing.py", line 259, in handle
    await self.app(scope, receive, send)
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/starlette/routing.py", line 61, in app
    response = await func(request)
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/fastapi/routing.py", line 226, in app
    raw_response = await run_endpoint_function(
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/fastapi/routing.py", line 161, in run_endpoint_function
    return await run_in_threadpool(dependant.call, **values)
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/starlette/concurrency.py", line 39, in run_in_threadpool
    return await anyio.to_thread.run_sync(func, *args)
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/anyio/to_thread.py", line 28, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(func, *args, cancellable=cancellable,
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 818, in run_sync_in_worker_thread
    return await future
  File "/home/andi/virtualenvs/aimaas/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 754, in run
    result = context.run(func, *args)
  File "/home/andi/workspace/aimaas/./backend/general_routes.py", line 242, in get_entity_change_details
    return traceability.entity_change_details(db=db, change_request_id=change_id)
  File "/home/andi/workspace/aimaas/./backend/traceability/entity.py", line 104, in entity_change_details
    _fill_in_field_change(change=change_, entity_change=change, entity=entity, listed_changes=listed_changes, checked_listed=checked_listed, db=db)
  File "/home/andi/workspace/aimaas/./backend/traceability/entity.py", line 44, in _fill_in_field_change
    change['changes'][attr.name] = {'new': v.new_value, 'old': v.old_value, 'current': get_old_value(db, entity, attr.name)}
  File "/home/andi/workspace/aimaas/./backend/traceability/entity.py", line 109, in get_old_value
    val = entity.get(attr_name, db)
  File "/home/andi/workspace/aimaas/./backend/models.py", line 144, in get
    raise KeyError(f'There is no attribute named `{attr_name}` defined for schema id {self.schema_id}')
KeyError: 'There is no attribute named `new` defined for schema id 16'
```
In order to keep historic data accessible we probably should not validate against the current attribute definitions :thinking:

commanderprice:
Well, we are not validating, we just show the current value to let user decide, whether the change still makes sense. I can just wrap call to `get_old_value` in try-except to set `current` to `None` if attribute is not found

crazyscientist:
>  I can just wrap call to `get_old_value` in try-except to set `current` to `None` if attribute is not found. Or we just can stop sending current values

I'm a bit undecided. For the existing change history listings on the schema/entity pages the `current` value does not provide too much benefit as it's very likely that the UI client has already fetched the current entity/schema.

With the new endpoint for pending change requests it might actually be a big benefit to have current values. 

Maybe we can introduce an API parameter to switch between the two options? The default could be to not send current values...

commanderprice:
> does not provide too much benefit

I'd rather say that it provides no benefit in this case🙃

> With the new endpoint for pending change requests it might actually be a big benefit to have current values.

Even before this endpoint it was good to have this data for pending requests; how this new route changes situation?🤔

> Maybe we can introduce an API parameter

We can, but should we?🤔

crazyscientist:
> Even before this endpoint it was good to have this data for pending requests; how this new route changes situation?

Currently I have a tabbed UI, so for users it's clear to which schema/entity the changes belong. And the schema/entity is already fetched. So one could just show this data.

My intention for the new endpoint is to provide a new top-level view that shows all pending change requests for all schemas and entities. When having only this list, I figured it would not be very client-friendly to require the client to fetch all the schemas/entities.

> We can, but should we?

I got the impression that the API responds pretty fast. I'm also fine with the `try ... except` solution :grin:

commanderprice:
> I'm also fine with the try ... except solution 

Then let's stick to it for now😎

crazyscientist:
Thank you for your work. This PR gets obsoleted by #36, #44 and #50.

PR #32, opened 2022-01-05
Basic Web UI

crazyscientist:
* Enabled CORS middleware on backend
* Added UI (based on Vue.js v3)

The UI elements already contain a number of features, that are not part of the backend, yet. These
elements have been visually disabled, but cause multiple "Not Found" error messages.

This commit is a collection of

* changes carved out from a bigger PR created by @commanderprice
* additional changes by @crazyscientist

PR #33, opened 2022-01-05
Container images (for backend and frontend) 

crazyscientist:
* Added `Dockerfile`s
* Added `Makefile` for automated builds
* Updated README

### Backend

* Use `psycopg2-binary` (instead of `psycopg2`)
* Only read env vars (instead of env_file)

### Frontend

* Handle different error format
* Pre-configured proxy for devel env


crazyscientist:
Caveat, this PR includes squashed commits from #32, which need to be dropped first...

crazyscientist:
Caveat, this PR uncovered #34 and contains a halfhearted workaround.

backend/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.9-slim
+
+COPY . /opt/aimaas
```
crazyscientist, line 3: Personally, I don't like installing Python packages this way, but on the other hand I did not want to refactor the project into a nice package structure when there are still pending changes. IMHO, this is the lesser of two evils.

crazyscientist (commented):
My five cents

PR #74, opened 2022-06-08
Freeze pydantic version due to test issues

crazyscientist:
pydantic 1.9.1 attempts to solve a memory leak, which causes a `KeyError`
in the test suite.


crazyscientist:
See https://github.com/samuelcolvin/pydantic/issues/4131

Issue #88, opened 2022-08-22
Handle deleted entities


crazyscientist:
In AIMAAS we do not delete entities (or schemas) from the database; we only mark them as deleted.

But there is currently no way to restore them.

* [x] The UI or API responses must indicate whether an entity/schema is deleted
* [x] There must be a way to undo the `deleted` marker
   * Either explicit: Undo deletion without additional data change
   * Or implicit: Undo deletion when trying to update entity/schema

Issue #69, opened 2022-03-08
Allow products to be set to confidential or internal


celan:
Some product data must not be published and therefore must be marked as "confidential". Other data is only of use for internal purposes and may therefore not be published.

PR #94, opened 2022-08-26
Visual indication for deleted schemas/entities (#88)

crazyscientist:
This commit adds indicators to components for showing details and
lists of entities/schemas.


frontend/src/components/EntityListTable.vue
```diff
@@ -18,7 +18,7 @@
       </tr>
       </thead>
       <tbody>
-      <tr v-for="e in entities" :key="e.id">
+      <tr v-for="e in entities" :key="e.id" :class="{'table-light': e.deleted}">
```
der-gabe, line 21: Forgive me, as I'm new to Vue.js, but I don't get this change... we're binding the class attribute (so far, so good), but not to a state or string, but to an to an object, which has `'table-light'` as its only property and the value of it is a boolean, indicating whether the entity `e` is marked as deleted or not... what does this do?

Does it somehow set `class="table-light"` if `e` is deleted?

crazyscientist: Yes, exactly. That is a [special case](https://vuejs.org/guide/essentials/class-and-style.html) of bindings.

frontend/src/components/SchemaList.vue
```diff
@@ -50,6 +51,12 @@
                 {{ schema.name }}
               </router-link>
             </div>
+            <div>
+              <i v-if="schema.deleted" class="eos-icons" data-bs-toggle="tooltip"
```
der-gabe, line 55: We probably don't need to leave an empty `div` here, if the schema is not marked as deleted.
```suggestion
            <div v-if="schema.deleted">
              <i class="eos-icons" data-bs-toggle="tooltip"
```

frontend/src/components/EntityListTable.vue
der-gabe: OK, good to know.

frontend/src/components/SchemaList.vue
crazyscientist: Right, currently not. With #69 this might change, but let's go with your suggestion until then.

der-gabe (approved):
Thanks! :+1:

der-gabe:
Approved, you can merge. Or should I merge, as the Assignee? I don't know what the process in this project is...

crazyscientist:
> you can merge. Or should I merge?

We can discuss the process, but I think its most efficient, when a reviewer performs the merge as part of the review. as I'm already on the page, I can perform the merge :slightly_smiling_face:

Issue #87, opened 2022-08-22
The `slug` field in not included in the search dialog


crazyscientist:
The page listing the entities of a schema offers to filter virtually all attributes/fields defined on the schema.

Only the field `slug` is missing.

crazyscientist:
Fixed by #96

PR #96, opened 2022-09-22
Case-insensitive search (fixes #95, #87)

### LINKED ISSUE #95: Case-insensitive search
The search should not be case-sensitive. For example, when I search for 'public' I get 0 hits but when I search for 'Public' I get the list of Public Cloud Modules.

crazyscientist:
* Add filter options for case-insensitive searching
* Also support filtering by entity name and slug
* Offer advanced filter options in the UI for name and slug

### DIFF
--- a/backend/crud.py
+++ b/backend/crud.py
@@ -427,7 +427,6 @@ def _parse_filters(filters: dict, attrs: List[str]) \
         else:
             attr_filters[attr][filter] = v
 
-    print("===DEBUG=== FILTERS", attr_filters, entity_filters)
     return attr_filters, entity_filters
 
 
--- a/backend/crud.py
+++ b/backend/crud.py
@@ -438,7 +438,7 @@ def _query_entity_with_filters(filters: dict, schema: Schema, all: bool = False,
     '''
     attrs = {i.attribute.name: i.attribute
              for i in schema.attr_defs if i.attribute.type.value.filters}
-    attrs_filters, entity_filters = _parse_filters(filters=filters, attrs=attrs.keys())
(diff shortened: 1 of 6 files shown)


der-gabe:
> * Also support filtering by entity name and slug
> * Offer advanced filter options in the UI for name and slug

Do we have an issue for that?

crazyscientist:
> Do we have an issue for that?

Nice catch. Yes, we had one: #87

backend/crud.py
```diff
@@ -437,18 +438,24 @@ def _query_entity_with_filters(filters: dict, schema: Schema, all: bool = False,
     '''
     attrs = {i.attribute.name: i.attribute
              for i in schema.attr_defs if i.attribute.type.value.filters}
-    attrs_filters, name_filters = _parse_filters(filters=filters, attrs=attrs.keys())
+    attrs_filters, entity_filters = _parse_filters(filters=filters, attrs=attrs.keys())
```
der-gabe, line 441: Above, this was renamed to `attr_filters` (singular first part).

While it's certainly not necessary, would it make sense to do the same here?

der-gabe (approved):
While I can't say with 100% certainty that this all does what it's supposed to, I went through all the changes and they all seem to make sense, to me. :+1:

backend/crud.py
crazyscientist: Yes, indeed. I had also considered changing it, but resisted the urge to refactor. But now that you mention it, let me get on it.

der-gabe:
Approved, you can merge. Or should I merge, as the Assignee? I don't know what the process in this project is...

crazyscientist:
Thank you

PR #105, opened 2022-10-19
Bugfixes for attributes of type "reference"

### LINKED ISSUE #79: Selection dialog for attributes of type "reference" not loading
If there is a schema defining an attribute of type "reference" the dialog for selecting an entity is not loading by default.

To be precise: Once the dialog is opened it's showing the spinner suggesting that entities are obtained from the API for the user to select, when in reality no API request was submitted. So, the user needs to manually click on the search icon to get options to choose from.

Instead, when opening the dialogs the API request should be sent.

### LINKED ISSUE #81: Attributes of type "reference" are not resolved in the list view
On schemas with attributes of type "reference" are not resolved (i.e. the progress indicator is shown) when the attribute is included in the list view.

This should be resolved.

crazyscientist:
* Load and show referenced entities in entity lists (fixes #81)
* Load the selction dialog in the edit form (fixes #79)

### DIFF
--- a/frontend/src/components/EntityList.vue
+++ b/frontend/src/components/EntityList.vue
@@ -65,6 +65,18 @@ export default {
       if (oldPage !== newPage) {
         this.getEntities({resetPage: false});
       }
+    },
+    'schema': {
+      async handler(newValue, oldValue) {
+        try {
+          if (newValue !== oldValue) {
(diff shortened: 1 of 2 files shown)


frozenIceage (approved):
Looks good to me, thanks!

PR #106, opened 2022-11-18
Bugfix for permission management of entities

crazyscientist:
A small bug allowed authenticated users to perform any actions on entities without permission, when the `reviewable` flag of the schema was unset. On the flip side, it prevented them from creating change requests for entities they had no permissions on.

This change fixes the situation:

* Any authenticated user is allowed to create change requests
* Only authorized users are allowed to apply changes to entities

### DIFF
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -11,7 +11,7 @@ from fastapi.testclient import TestClient
 from ..auth import authenticated_user, authorized_user
 from ..auth.crud import get_or_create_user, get_user, grant_permission
 from ..auth.enum import RecipientType, PermissionType
-from ..auth.models import User
+from ..auth.models import User, Permission
 from ..database import get_db
 from ..models import *
 from ..schemas.auth import UserCreateSchema, PermissionSchema
@@ -351,6 +351,22 @@ class OldStyleTestClient(TestClient):
             extensions=extensions,
             json=json
         )
+
+
+@pytest.fixture
+def unauthorized_testuser(dbsession, testuser) -> User:
+    # The testuser is a superuser. Remove permissions.
+    dbsession.query(Permission).filter(Permission.recipient_type == RecipientType.USER,
+                                       Permission.recipient_id == testuser.id).delete()
+    return testuser
+
+
+@pytest.fixture
+def unauthorized_testuser(dbsession, testuser) -> User:
+    # The testuser is a superuser. Remove permissions.
+    dbsession.query(Permission).filter(Permission.recipient_type == RecipientType.USER,
+                                       Permission.recipient_id == testuser.id).delete()
+    return testuser
 
 
 @pytest.fixture
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -362,14 +362,6 @@ def unauthorized_testuser(dbsession, testuser) -> User:
 
 
 @pytest.fixture
-def unauthorized_testuser(dbsession, testuser) -> User:
-    # The testuser is a superuser. Remove permissions.
-    dbsession.query(Permission).filter(Permission.recipient_type == RecipientType.USER,
-                                       Permission.recipient_id == testuser.id).delete()
-    return testuser
-
-
-@pytest.fixture
 def client(dbsession):
     app = create_app(session=dbsession)
     app.dependency_overrides[get_db] = lambda: dbsession
--- a/backend/tests/test_auth_api.py
+++ b/backend/tests/test_auth_api.py
@@ -395,6 +395,9 @@ class TestRoutesRequiringAuth:
 
     def test_approve_request(self, dbsession: Session, unauthorized_testuser: User,
                              authenticated_client: TestClient):
+        """
+        Test that an authenticated but unauthorized user is not allowed to approve a change request.
+        """
         change_request_ids = dbsession.query(ChangeRequest.id)\
             .where(ChangeRequest.status == ChangeStatus.PENDING)\
             .first()
--- a/backend/tests/test_auth_api.py
+++ b/backend/tests/test_auth_api.py
@@ -12,6 +12,8 @@ from ..config import settings
 from ..models import Schema, Entity
 from ..schemas import BaseGroupSchema, GroupSchema, UserCreateSchema, PermissionSchema, \
     RequirePermission
+from ..traceability.enum import ReviewResult, ChangeStatus
+from ..traceability.models import ChangeRequest
 from .conftest import TEST_USER
 from .mixins import CreateMixin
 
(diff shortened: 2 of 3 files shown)


backend/tests/conftest.py
```diff
@@ -353,6 +353,22 @@ def delete(self, url, *, params=None, headers=None, cookies=None, auth=USE_CLIEN
...
+
+
+@pytest.fixture
+def unauthorized_testuser(dbsession, testuser) -> User:
+    # The testuser is a superuser. Remove permissions.
+    dbsession.query(Permission).filter(Permission.recipient_type == RecipientType.USER,
+                                       Permission.recipient_id == testuser.id).delete()
+    return testuser
+
+
```
der-gabe, line 371: Why is this here twice, i.e. doubled?

Looks identical to me…

Oversight, or am I missing something?

backend/tests/test_auth_api.py
```diff
@@ -385,12 +385,61 @@ def test_require_authentication(self, method, url, client: TestClient):
...
+            .join(Schema)\
+            .filter(Schema.slug == "unperson", Entity.slug == "foo")\
+            .exists()
+        assert not dbsession.query(Entity.id).filter(q).scalar()
+
+    def test_review_change_request_breakout__update(self, dbsession: Session,
+                                                    unauthorized_testuser: User,
+                                                    authenticated_client: TestClient):
+        """
+        Test that authenticated user is allowed to create a change request without actually changing
```
der-gabe, line 414:
```suggestion
        Test that authenticated user is allowed to update a change request without actually changing
```

backend/tests/test_auth_api.py
```diff
@@ -385,12 +385,61 @@ def test_require_authentication(self, method, url, client: TestClient):
...
+
+        assert response.status_code == 202
+        dbsession.refresh(entity)
+        assert entity.name == "Foo"
+
+    def test_review_change_request_breakout__delete(self, dbsession: Session,
+                                                    unauthorized_testuser: User,
+                                                    authenticated_client: TestClient):
+        """
+        Test that authenticated user is allowed to create a change request without actually deleting
```
der-gabe, line 432:
```suggestion
        Test that authenticated user is allowed to delete a change request without actually deleting
```

backend/tests/conftest.py
crazyscientist: It should suffice to say: Error between keyboard and chair. Let me delete the duplication...

backend/tests/test_auth_api.py
```diff
@@ -385,12 +385,61 @@ def test_require_authentication(self, method, url, client: TestClient):
...
+        an entity
+        """
+        schema = dbsession.query(Schema).where(Schema.slug == "unperson").one()
+        entity = Entity(name="Foo", slug="foo", schema_id=schema.id)
+        dbsession.add(entity)
+        dbsession.commit()
+        response = authenticated_client.delete("/entity/unperson/foo",
+                                               json={"name": "Føø", "slug": "foo"})
```
der-gabe, line 440: Do we even need the `json` argument in this case?

And if so, shouldn't it match the original entity's values?

backend/tests/test_auth_api.py
```diff
@@ -385,12 +385,61 @@ def test_require_authentication(self, method, url, client: TestClient):
...
 
     @pytest.mark.parametrize("method, url", authorized)
-    def test_require_authorization(self, method, url, dbsession: Session, testuser: User,
-                                   authenticated_client: TestClient):
-        # The testuser is a superuser. Remove permissions first.
-        dbsession.query(Permission).filter(Permission.recipient_type == RecipientType.USER,
-                                           Permission.recipient_id == testuser.id).delete()
-
+    def test_require_authorization(self, method, url, dbsession: Session,
+                                   unauthorized_testuser: User, authenticated_client: TestClient):
```
der-gabe, line 389: This test has only been refactored, bu it didn't catch the original bug before, so wouldn't one or a few new tests be in order?

backend/tests/test_auth_api.py
crazyscientist: Right, that's what you get for working late in the evening...

```suggestion
        response = authenticated_client.delete("/entity/unperson/foo")
```

backend/tests/test_auth_api.py
crazyscientist: Yes, that's what the new `test_review_change_request_breakout_*` are for :slightly_smiling_face: 

We don't want an unauthorized user to change any entity data, but they should be allowed to submit a request

crazyscientist: But we have not covered the case of an unauthorized user accepting the request. Let me add that...

backend/tests/test_auth_api.py
```diff
@@ -385,12 +387,72 @@ def test_require_authentication(self, method, url, client: TestClient):
...
         func = getattr(authenticated_client, method)
         response = func(url)
         assert response.status_code == 403
+
+    def test_approve_request(self, dbsession: Session, unauthorized_testuser: User,
+                             authenticated_client: TestClient):
+        change_request_ids = dbsession.query(ChangeRequest.id)\
```
der-gabe, line 398:
```suggestion
        """
        Test that an authenticated but unauthorized user is not allowed to approve a change request.
        """
        change_request_ids = dbsession.query(ChangeRequest.id)\
```

backend/tests/test_auth_api.py
der-gabe: That's great, but my point was that apparently, if I understand the issue correctly, authenticated, but unauthorized users were able to make any kind of change they wanted – directly, i.e. w/o a change request.

Is that correct?

Because if so, then why wasn't *that* caught by any test?

The new tests merely check that such users can create change _requests_ (which themselves don't change anything directly), but isn't there also a way to make changes w/o requests? And if so, where is there a test that unauthorized users cannot make such changes?

crazyscientist: > why wasn't that caught by any test?

Because the existing test (intended to verify permissions on direct entity changed) checked against the wrong schema (i.e. the one which required reviews). So, the test incorrectly checked that unauthorized users were not allowed to create a change request. By changing to the schema with an unset `reviewable` flag, the test checks what is actually intended: Unauthorized users are not allowed to change an entity.

The new tests cover the other way of doing things: Unauthorized users may create change requests, when the `reviewable` flag is set, but not apply them to the entity.

Does that clarify things?

der-gabe: > Because the existing test […]

i.e. `test_require_authorization`, right?

> […] checked against the wrong schema (i.e. the one which required reviews). So, the test incorrectly checked that unauthorized users were not allowed to create a change request. By changing to the schema with an unset `reviewable` flag, the test checks what is actually intended: Unauthorized users are not allowed to change an entity.

This implies that the existing **test** should have been _functionally_ changed in some way (to test the right schema), but it _hasn't!_ The only thing that has changed is that some setup code has been moved into a fixture. Great, but that's only a refactoring.

The test used to pass, when the bug was present, and it hasn't functionally changed.

Why should I now have confidence in a test that previously did not fail in the presence of the exact kind of bug it's meant to detect?

If I were to reintroduce the same bug again, would `test_require_authorization` now fail?

If so, I don't understand why.

If not, then there's still at least one test missing.

I'm open to having a call about this, but ATM it's super unclear.

> The new tests cover the other way of doing things: […]

That part is clear to me and well done for adding those tests.

crazyscientist: Yes, we are talking about `test_require_authorization`, which is also a parameterized test. In the parameter definition the schema was changed from `unperson` to `person`.

If you were to reintroduce the bug, the testsuite would fail as expected.

If you like, give me a ring...

der-gabe: Ah... :bulb: the parametrisation changed!

I meant to ask about that change from `unperson` to `person` and how it ties in, but then I forgot.

Now I see it and it makes sense.

der-gabe (approved):

crazyscientist:

PR #108, opened 2022-11-24
Fixed problem in mixin for testsuite

crazyscientist:
When granting permissions on a specific object type and a specific object ID, the ID must belong to an object of this type!


der-gabe:
LGTM

der-gabe (approved):

PR #111, opened 2022-12-15
Fixed permission handling for creation of entities

crazyscientist:
* In `route_create_entity` and required permission was not specifying the schema ID.
* In `authorized_user` the ID or slug of the schema was assumed to be passed as path param.

Both contributed to authorized users (who were no superusers) being denied to create entities. This commit now enables them to do so.

### DIFF
--- a/backend/auth/__init__.py
+++ b/backend/auth/__init__.py
@@ -62,19 +62,25 @@ def authorized_user(required_permission: RequirePermission) -> Callable:
     async def is_authorized(request: Request, db: Session = Depends(get_db),
                             user: User = Depends(authenticated_user)) -> User:
         if isinstance(required_permission.target, (Schema, Entity)):
-            id_or_slug = request.path_params.get("id_or_slug", None)
-            try:
-                required_permission.target.id = int(id_or_slug)
-            except (ValueError, TypeError):
-                Model = required_permission.target.__class__
-                try:
-                    required_permission.target.id = db.query(Model.id) \
-                                                      .filter(Model.slug == id_or_slug).one()[0]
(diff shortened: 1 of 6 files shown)


der-gabe:
Considering this together with PRs #106 and #108 it seems that permission handling is, if not the Achilles' heel, at least a weak point of AIMAAS as it stands today.

That may be OK for an internal tool, but it's not great for a tool that may one day be exposed to the wider world.

I'm wondering whether we should address this in some way. Maybe through a dedicated review of the security model and its implementation, but that's only one idea. What do you think @crazyscientist?

crazyscientist:
> Considering this together with PRs it seems that permission handling is, if not the Achilles' heel, at least a weak point of AIMAAS as it stands today.

Yes, this is now the second "small" bug in this particular area with some serious impact. The other PR you mentioned does not address a problem with permission handling, it addresses an incorrect fixture, which happens to be a permission.

> Maybe through a dedicated review of the security model and its implementation

It certainly wound qualify as "better to be safe than sorry".

der-gabe:
>     * In `authorized_user` the ID or slug of the schema was assumed to be passed as path param.

Also, the old logic would always try to overwrite the required permission's target ID, even when one was already present. That was a safe bet when coming from `route_create_entity` (and that's no longer the case, thanks to the other fix in this PR), but I'm not so sure about other code paths. Either way, it now checks whether there's already an ID there and leaves it alone, if so.

der-gabe (approved):
LGTM! :slightly_smiling_face: :+1:

crazyscientist:

PR #123, opened 2023-02-20
New view for bulk editing of entities (closes #90)

### LINKED ISSUE #90: Allow bulk change of dates
Sometimes a set of modules or products have a change of lifecycle dates. Without a bulk change that can be tedious.

Use case:
Chris User needs to set a Beta product as released by removing the word "beta" from all affected products and wants to set the actual FCS date, which is not known before it happens.

crazyscientist:
* Added new Vue route
* Added a new action button to entity list (to go to new view)
* Fixed issue with non-unique IDs in Checkbox.vue
* Applied changes to EntityForm.vue to enable use in parallel
* Removed obsolete debug output

### DIFF
--- a/frontend/src/components/EntityBulkEdit.vue
+++ b/frontend/src/components/EntityBulkEdit.vue
@@ -1,0 +1,87 @@
+<template>
+  <BaseLayout>
+    <template v-slot:additional_breadcrumbs>
+      <li class="breadcrumb-item active">
+        <router-link :to="{name: 'schema-view', params: {schemaSlug: activeSchema?.slug || 'n-a'}}">
+          {{ activeSchema?.name || 'n/a' }}
+        </router-link>
+      </li>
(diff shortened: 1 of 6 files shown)


crazyscientist:
The following screenshots show the changes to the entity listing and the new view for bulk editing

grafik

grafik

s-leon:
If you select entries in the `EntityList` and then switch pages the selection footer stays visible with the selected Entries still visible and actionable.
However, if one then selects any Entry on the current page the previous selection gets overwritten.
From a standpoint of consistency it would be preferable to allow adding to the batch regardless of page, or clear and hide the `EditForm`

crazyscientist:
Nice catch. But this glitch is unrelated to this new feature, so I created a separate PR for that: #133

s-leon (approved):
LGTM

Issue #98, opened 2022-09-28
Error message "Not authenticated" is too small and low


celan:
Especially on high resolutions the error message is barely visible. Maybe it could be displayed next or above the form?

PR #124, opened 2023-03-10
Move alert display to top of screen and enlarge text

r-richardson:
Regarding [Issue 98](https://github.com/SUSE/aimaas/issues/98)
This PR enhances alert visibility by moving the alert display from the bottom right to the top right corner of the screen and enlarging the text displayed within alerts by 50%.

### DIFF
--- a/frontend/src/components/alerts/AlertDisplay.vue
+++ b/frontend/src/components/alerts/AlertDisplay.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="container-fluid fixed-bottom m-3 d-flex flex-column align-items-end"
-       id="alert-contaier">
+       id="alert-container">
       <div v-for="alert of alerts" :key="alert.id"
            class="alert shadow alert-dismissible fade show w-50 me-3"
            :class="`alert-${alert.level}`" role="alert">
@@ -58,7 +58,7 @@ export default {
 </script>
 
 <style scoped>
-#alert-contaier {
+#alert-container {
   z-index: 10;
 }
 </style>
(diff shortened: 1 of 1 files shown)


frontend/src/components/alerts/AlertDisplay.vue
```diff
@@ -1,6 +1,6 @@
 <template>
-  <div class="container-fluid fixed-bottom m-3 d-flex flex-column align-items-end"
-       id="alert-contaier">
+  <div class="container-fluid fixed-top m-3 d-flex flex-column align-items-end"
+       id="alert-container">
```
crazyscientist, line 3: Thank you very much for fixing this typo :+1:

frontend/src/components/alerts/AlertDisplay.vue
```diff
@@ -58,7 +58,8 @@ export default {
 </script>
 
 <style scoped>
-#alert-contaier {
+#alert-container {
   z-index: 10;
+  font-size: 1.5em;
```
crazyscientist, line 63: This is certainly a correct solution. Maybe I can interest you in a less explicit approach?

Instead of directly defining the CSS, we could add [`fs-4`](https://getbootstrap.com/docs/5.3/utilities/text/#font-size) to the classes of `alert-container`. Would you like to give it a try?

r-richardson: Ah yes thanks, i was looking for something like that, so that is then using bootstrap if i'm correct ?

crazyscientist: Yes, bootstrap5 with the SUSE theme: https://github.com/crazyscientist/suse-bootstrap5-theme

crazyscientist (approved):
LGTM 👍

PR #126, opened 2023-03-20
div blocking navbar clicks (from #98)

r-richardson:
Fixes blocking of input while alerts are being displayed. Bug introduced by [PR #124]


frontend/src/components/alerts/AlertDisplay.vue
```diff
@@ -1,8 +1,8 @@
 <template>
-  <div class="container-fluid fixed-top m-3 d-flex flex-column align-items-end fs-4"
+  <div class="container-fluid fixed-top m-3 d-flex flex-column w-50 fs-4 ms-auto"
```
crazyscientist, line 2:
```suggestion
  <div class="container-fluid fixed-top m-3 d-flex flex-column w-50 fs-4 ms-auto p-0"
```

frontend/src/components/alerts/AlertDisplay.vue
```diff
@@ -1,8 +1,8 @@
...
        id="alert-container">
       <div v-for="alert of alerts" :key="alert.id"
-           class="alert shadow alert-dismissible fade show w-50 me-3"
+           class="alert shadow alert-dismissible fade show me-3"
```
crazyscientist, line 5:
```suggestion
           class="alert shadow alert-dismissible fade show"
```

crazyscientist (changes requested):
Nice catch and thank you for the fix :slightly_smiling_face: 

I only have two small suggestions to make the gaps between the alerts and the top and right window border the same size.

crazyscientist (approved):

Issue #75, opened 2022-07-08
Set a speaking window title


tgoettlicher:
Currently the window title says "frontend". A more speaking window title would be "SUSE Product Lifecycle Database".

r-richardson:
I couldnt find a window titled "frontend", neither locally nor in production. It shows "AIMAAS", no matter what page.
I'd say "AIMAAS" is a good default title, but maybe we should still change it to "SUSE Product Lifecycle Database" for production ?

crazyscientist:
Ah, right. A while back I replaced "frontend" with "AIMAAS" (see 29ad8fb0134b3aa101213052eb73cbdb0f461156)

I would not change the hard-coded package name. What we could consider is to adjust the page title (in the browser) depending on the currently active view :thinking:

PR #127, opened 2023-03-22
Add dynamic window titles (from #75)

r-richardson:
**WIP**
as discussed in #75, this PR adds a `usePageTitle.js` composable which allows for dynamic and reactive window titles, also checks the routes meta information (defined in `router.js`) for predefined `meta.title` variables and uses these if no dynamic title was passed. If no explicit title was passed nor found within the router.js routes definitions, the default title `AIMAAS` is used instead

### DIFF
--- a/frontend/src/composables/usePageTitle.js
+++ b/frontend/src/composables/usePageTitle.js
@@ -1,0 +1,18 @@
+import { ref, onMounted, watch } from 'vue';
+import { useRoute } from 'vue-router';
+
+export default function usePageTitle(pageTitle) {
+  const defaultTitle = "AIMAAS"; // used when no title is defined for given route
+  const route = useRoute();
+  const title = ref(defaultTitle);
+
+  onMounted(() => { // used on first load to DOM
+    title.value = pageTitle || route.meta.title || defaultTitle;
+    document.title = title.value;
+  });
+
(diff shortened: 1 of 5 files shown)


frontend/src/composables/usePageTitle.js
```diff
@@ -0,0 +1,22 @@
...
+import { useRoute } from 'vue-router';
+
+export default function usePageTitle(pageTitle) {
+  const defaultTitle = "AIMAAS"; // used when no title is defined for given route
+  const route = useRoute();
+  const title = ref(defaultTitle);
+
+  if (pageTitle) {
+  document.title = pageTitle || route.meta.title || defaultTitle;
+}
```
crazyscientist, line 11: I think, these lines could be superfluous. If I comment them out, it still works as expected.

```suggestion
```

crazyscientist (changes requested):
Thank you. I only have one small suggestion for changes :+1:

frontend/src/composables/usePageTitle.js
crazyscientist: Thank you for changing this.

On a side note: You did not have to rebase and force-push. An additional commit (as created when clicking on "Commit suggestion" would have been fine, too. :slightly_smiling_face:

crazyscientist (approved):

PR #130, opened 2023-04-04
Upgraded NPM packages

crazyscientist:


s-leon:
1. So instantly NPM notifies about a `high severity vulnerability` in webpack.
If we are already upgrading packages we could update that version, too.

2. running it results in the message: `caniuse-lite is outdated` [reference](https://github.com/browserslist/update-db#readme) 

   while I'm not really familiar with it, it seems this is also a housekeeping item to be done if one already updates the `package-lock`

crazyscientist:
I keep forgetting to run `npm up` after `ncu -u` :grimacing:

Thanks for the reminder :+1:

PR #133, opened 2023-04-20
Fixed stale references on array with selected entities

crazyscientist:
When a new page with entity data is loaded the `getEntities` assigns a new Arrays to its `selected` attribute, which breaks existing references.

Also, the selection should be cleared. However, in the future somebody might prefer to support selections across multiple pages; but not for now.


s-leon (approved):
LGTM

PR #135, opened 2023-04-21
Encode credentials correctly in login request (fixes #134)

### LINKED ISSUE #134: Icorrect encoding of request body in `plugins.api.API.login`
The body for the login request is constructed like this:

```javascript
        const body = `username=${username}&password=${password}`;
        const response = await this._fetch({
            url: url,
            method: 'POST',
            body: encodeURI(body),
            headers: {'Content-Type': 'application/x-www-form-urlencoded'}
        });
```

This is incorrect and prevents users from logging in, if the password starts with a `$`.

crazyscientist:


frontend/src/plugins/api.js
```diff
@@ -281,11 +281,14 @@ class API {
 
     async login({username, password} = {}) {
         const url = `${this.base}/login`;
-        const body = `username=${username}&password=${password}`;
+        const body = new URLSearchParams({
+            username: username,
```
s-leon, line 285: Yes from testing it, it works (even with + in the password)...

But when I looked at the [docs](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams?retiredLocale=de#preserving_plus_signs), it suggested using `encodeURIComponent()`

When I look at the result it works and is correctly encoded, so no objections but still a bit confused by the docs

s-leon (approved):
LGTM

Issue #138, opened 2023-05-23
Define an appropriate timeout for `fetch` calls in the frontend


crazyscientist:
By default, `fetch` uses a timeout of 300 seconds, which can be longer than appropriate.

In `api.js` requests should time out more quickly. 10 seconds sounds like a more appropriate value.

References:

* https://developer.mozilla.org/en-US/docs/Web/API/AbortController
* https://dmitripavlutin.com/timeout-fetch-request/
* https://dmitripavlutin.com/timeout-fetch-request/

crazyscientist:
Fixed by #140

PR #139, opened 2023-05-23
added configurable timeout to the `_fetch` function #138

esskayesss:
this merge request adds a configurable timeout to the `_fetch` function in `api.js`. by default, the function now uses a timeout of 10 seconds, as suggested in the issue, and also seems to be more appropriate for most requests. the timeout can be overridden by passing a timeout parameter to the function.

this change addresses issue #138, which requested that the fetch calls should time out more quickly. the change makes the code more flexible and allows it to be used in a wider range of scenarios.

please review and merge this change if it meets the project's standards.

### DIFF
--- a/frontend/src/plugins/api.js
+++ b/frontend/src/plugins/api.js
@@ -89,7 +89,7 @@ class API {
         }
     }
 
-    async _fetch({url, headers, body, method} = {}) {
+    async _fetch({url, headers, body, method, timeout = 10000} = {}) {
         let allheaders = {"Content-Type": "application/json"};
         const token = this.token;
         if (token) {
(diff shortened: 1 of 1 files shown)


crazyscientist:
Thank you for submitting this solution so quickly.

Out of curiosity: What error message will be shown, if a request times out? (I'm only asking because I'm not at a computer where I could test it right away)

esskayesss:
no, you're totally right in raising this. i will give it more thought and create a new pull request keeping all the dealings locally. i'm closing this for now. i'll try to create a new one by tomorrow. do you have any suggestions as to how the timeout should be handled?

PR #140, opened 2023-05-25
added configurable timeout to `_fetch()` #138

esskayesss:
this is a follow-up to #139
this change addresses issue #138, which requested that the fetch calls should time out more quickly. the change makes the code more flexible and allows it to be used in a wider range of scenarios.

this merge request introduces a flexible timeout option for the `_fetch` function in `api.js`. the function now defaults to a timeout of *10* seconds, which is in line with the issue recommendation and suitable for most requests. the timeout can be adjusted by passing a timeout argument to the function.

> If you would like to dig deeper and provide a nice error message to a user ... It should be sufficient for the message to simply state that an operation timed out.

as rightfully suggested by @crazyscientist, 
the `AbortError` no longer clutters the screen or the console. i have invoked the `_error_to_alert()` function to alert the user of the timeout. additionally, more information is logged in the console for the devs.

### DIFF
--- a/frontend/src/plugins/api.js
+++ b/frontend/src/plugins/api.js
@@ -89,7 +89,7 @@ class API {
         }
     }
 
-    async _fetch({url, headers, body, method} = {}) {
+    async _fetch({url, headers, body, method, timeout_s = 10000} = {}) {
         let allheaders = {"Content-Type": "application/json"};
         const token = this.token;
         if (token) {
@@ -104,15 +104,26 @@ class API {
         }
 
         try {
+            const controller = new AbortController();
+            const timeout = setTimeout(() => controller.abort(), timeout_s);
+
             const response = await fetch(url, {
                 method: method || 'GET',
                 body: encoded_body,
-                headers: allheaders
+                headers: allheaders,
+                signal: controller.signal,
             });
+
+            clearTimeout(timeout);
             await this._is_response_ok(response);
             return await response.json();
         } catch (e) {
-            this._error_to_alert(e);
+            if(e.name === "AbortError") {
+                this._error_to_alert({message: "Request timed out"});
+                console.error("request timed out: %s", url);
+            } else {
+                this._error_to_alert(e);
+            }
             return null;
         }
     }


frontend/src/plugins/api.js
```diff
@@ -104,15 +104,26 @@ class API {
...
             await this._is_response_ok(response);
             return await response.json();
         } catch (e) {
-            this._error_to_alert(e);
+            if(e.name === "AbortError") {
+                this._error_to_alert({message: "Request timed out"});
+                console.error("request timed out: %s", url);
+            } else {
+                this._error_to_alert(e);
+            }
```
crazyscientist, line 126: Thank you very much for your contribution and my apologies for not picking this up earlier.

In principle, the solution looks perfectly fine. I'd only like to ask you small favor of you: Would you mind moving the handling of `AbortError` into the `_error_to_alert` method?

crazyscientist (changes requested):

frontend/src/plugins/api.js
esskayesss: i was considering this approach but it had some drawbacks:
- the `AbortError` that this code produces is quite _ugly_ and does not show that there was a request timeout. this is a data point that the user should be notified about, instead of expecting them to read a long message, or even make sense of the word `AbortError`.
- considering this issue, if i display a message saying that a request timed out in the `_error_to_alert` function, it could be misleading if the code were to throw an `AbortError` for another reason.

so i'd request some feedback on this before i proceed to implement it.

crazyscientist: Yes, those are very good arguments.

I had a quick look at the docs of [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort), which gave me another idea:

How about, we leave the content of the `catch` block unchanged and instead use your self-defined object as argument to the call to the `abort()` method in line 108?

E.g.:
```js
            const timeout = setTimeout(() => controller.abort(
              {"message": "Request timed out."}), 
              timeout_s
            );

Comment (esskayesss) [reply frontend/src/plugins/api.js:126]: apparently, the `controller.abort(reason)` doesn't make it throw the reason object or string. it just sets the value of `controller.signal.reason` instead.

for example, the following code:
```js
async function delay (ms, value) {
  return new Promise(res => setTimeout(() => res(value), ms));
}

async function example () {
  const controller = new AbortController();
  try{
		const timeout = setTimeout(() => controller.abort(
      {"message": "Request timed out"}), 1);
    const response = await fetch("https://google.com", {signal: controller.signal});
		await delay(2000);
    console.log(controller.signal.reason);
	} catch (e) {  
    console.log(e.constructor);
    console.log(controller);
    console.log(controller.signal);
	}
}

console.log("calling method");
example();
```

produces this output in the console:
```
"calling method"
function DOMException() { [native code] }
[object AbortController] {
  abort: function abort() { [native code] },
  signal: [object AbortSignal] {
    aborted: true,
    addEventListener: function addEventListener() { [native code] },
    dispatchEvent: function dispatchEvent() { [native code] },
    onabort: null,
    reason: { ... },
    removeEventListener: function removeEventListener() { [native code] },
    throwIfAborted: function throwIfAborted() { [native code] }
  }
}
[object AbortSignal] {
  aborted: true,
  addEventListener: function addEventListener() { [native code] },
  dispatchEvent: function dispatchEvent() { [native code] },
  onabort: null,
  reason: {
    message: "Request timed out"
  },
  removeEventListener: function removeEventListener() { [native code] },
  throwIfAborted: function throwIfAborted() { [native code] }
}
```

`throwIfAborted` is interesting, but it will require a lot more tinkering.

crazyscientist: To be honest, I had not bothered to double check the implementation of `AbortController`, but when I tested it in Firefox, the catched object `e` was identical to the object passed into `abort()`.

In my case, this code:

```js
        const reason = {message: "Request timed out"};
        try {
            const controller = new AbortController();
            const timeout = setTimeout(() => controller.abort(reason), timeout_s);
            // ...
        } catch (e) {
            console.log("Time out", e, typeof e, e === reason)
            this._error_to_alert(e);
            return null;
        }
```
produces this output on the JS console of Firefox:
grafik

---

And now for the fun part: Chrome behaves differently, here `e` is a `DOMException`, which does not contain teh reason. :face_with_head_bandage: 

With that in mind, my proposal really is not alternative.

crazyscientist (approved):

frontend/src/plugins/api.js
esskayesss: it's interesting that it ran on Firefox lol.

crazyscientist: Indeed, one might even wonder whether this is the beginning of another mess similar to Internet Explorer :rofl:

PR #144, opened 2023-06-01
Changed assertions in `backend/tests/test_crud_schema.py::TestSchemaRead::test_get_schemas`

crazyscientist:
The success of the test cases depended on the default order in which rows are returned by the database. With this change, this dependency is resolved.


s-leon (approved):
LGTM :+1:

PR #145, opened 2023-06-01
Removed version constraint of `pydantic`

crazyscientist:
The constraint was introduced in #74 due to https://github.com/pydantic/pydantic/issues/4131. The fix for pydantic has been released and the constraint is now superfluous.

### DIFF
--- a/backend/main.py
+++ b/backend/main.py
@@ -1,6 +1,9 @@
+from fastapi_pagination import add_pagination
+
 from . import create_app, models
 from .database import engine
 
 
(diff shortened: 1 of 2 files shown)


crazyscientist:
See https://github.com/pydantic/pydantic/issues/4131#issuecomment-1571880345:

> I'd still recommend using a constraint pydantic<2 in preparation for Pydantic v2 release. There will be some significant breaking changes in v2

s-leon (commented):
LGTM

s-leon (approved):

PR #152, opened 2023-08-23
Fix superuser group inheritance

### LINKED ISSUE #149: The `SUPERUSER` permission set on groups has no effect
In the function `auth.crud.has_permission` is a bug. 

It only considers the `PermissionType.SUPERUSER` granted directly to users. If the permission is inherited from a group, it has no effect.

r-richardson:
- Fix #149 
- Add test case checking if inherited superuser privileges take effect.


crazyscientist (approved):
Thank you for fixing this :slightly_smiling_face:

PR #155, opened 2023-09-22
Fixed bug in creation of new entity with list attr.

### LINKED ISSUE #148: Creation of an entity with list attributes is not possible
# Description

In the dialog to create a new entity, one cannot add items to a list attribute due a JS exception: `Uncaught (in promise) TypeError: $props.entity is undefined` in EntityForm.vue.

# Steps to reproduce

1. Create a schema with a multi-valued string attribute (aka. List button)
2. Go to the "Add Entity" form of this schema
3. Click on the "Add" button of that multi-valued attribute

No input field appears, but instead a JS exception.

# Expected behavior

Clicking on "Add" adds an input element that can be edited.

crazyscientist:
The EntityForm component is also used for creating new entities, so it cannot rely on entity IDs to be available.

This closes #148.

### DIFF
--- a/frontend/src/components/inputs/EntityForm.vue
+++ b/frontend/src/components/inputs/EntityForm.vue
@@ -29,7 +29,7 @@
             <div class="col-lg-10">
               <ul class="list-group">
                 <li v-for="(val, idx) in (editEntity[attr.name] || [])"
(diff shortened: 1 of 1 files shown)


s-leon (approved):
works now, LGTM

PR #161, opened 2023-10-18
Build and publish container images

crazyscientist:
* Use Python 3.12
* Moved `alembic.ini` (to include it in image)
* Allow different path to `alembic.ini` in test suite via env. variable
* Added an action to build, publish and test images
* Added an action to delete old images


.github/workflows/housekeeping.yml
```diff
@@ -0,0 +1,20 @@
...
+    - cron: "0 0 1 * *"
+jobs:
+  clean-ghcr:
+    name: Delete old container images
+    runs-on: ubuntu-latest
+    steps:
+      - name: Delete old releases
+        uses: snok/container-retention-policy@v2
+        with:
+          image-names: aimaas/aimaas-api, aimaas/aimaas-ui
```
der-gabe, line 13: For consistency with how we do it in `on_tag.yml`?
```suggestion
          image-names: ${GITHUB_REPOSITORY,,}/aimaas-api, ${GITHUB_REPOSITORY,,}/aimaas-ui
```
Although I have to admit that I don't know what the two commas in the curly braces do...

der-gabe (changes requested):
2 tiny remarks. LGTM, otherwise.

.github/workflows/housekeeping.yml
crazyscientist: > what the two commas in the curly braces do

Bash parameter expansion which converts the variable name to lower case.

der-gabe (approved):

Issue #178, opened 2023-12-18
Need to reload almost all time


jsrain:
After virtually any change (adding a use case, assigning a use case to a package) one needs to reload the browser. Otherwise, new use case is not shown, random package shows assigned to use case of different package, etc. This makes the whole UI almost unusable.

crazyscientist:
Related to #157

PR #187, opened 2024-01-18
Refresh entity detail and schema detail on update

TeodoraPavlova:
Related to https://github.com/SUSE/aimaas/issues/178

Applied fixes for:

1. When you edit an entity/attributes in entity detail page and you click save, the values of attributes of type FK are vanishing. Previously manual page reload showed them again and now it's not needed.

2. If you go back from entity detail page to schema details, entity list is not refreshed. Now it is.

### DIFF
--- a/frontend/src/components/EntityList.vue
+++ b/frontend/src/components/EntityList.vue
@@ -75,6 +75,9 @@ export default {
       default: false
     }
   },
+  activated() {
+    this.getEntities({resetPage: false});
+  },
   computed: {
     pages: computed(() => {
       try {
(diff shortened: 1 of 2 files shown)


crazyscientist (approved):
I believe I understand what you meant in yesterday's standup.

The fact that we need to call `getSelected` after `this.selected` was modified by `getSelected`, `onSelect` or `onClear` (indirectly via triggering a change of the `modelValue` param) looks superfluous.

To avoid such suboptimal solutions we would need a large scale review and refactoring of the UI, which clearly is out of scope of #178.

Thank you for finding this simple fix.

PR #189, opened 2024-01-20
Allow sorting of entities by attribute of type reference

frozenIceage:
* frontend: allow sending sort by reference
* backend: add correlated subquery in order by clause


crazyscientist (approved):

Issue #191, opened 2024-01-22
Showing package catalog details does not properly show the data set


dirkmueller:
There seems to be lack of refresh for "use case" field when browsing from one that had a use case set to another one that has no use case set

steps: go to

https://aimaas.suse.de/schema/package-catalog/suse-sles-15-sp5-OpenPrintingPPDs

ctrl-R refresh one or more times. it shows "Printing Drivers" as usecase, which is correct. 

Then go to

https://aimaas.suse.de/schema/package-catalog/suse-sles-15-sp5-hp-drive-guard

and it shows "Printing drivers" also. After ctrl-R refresh, the use case is emptied (which is correct as no use case has been set)

crazyscientist:
# Analysis

When following the steps by entering the URLs in the browser address bar, this cannot be reproduced. The important part for reproduction is to go back to the entity list:

1. Navigate to entity list.
2. Follow link to an entity E1 (with a set attribute of type reference). The list of referenced entities gets displayed correctly.
3. Use the "Back" button of the browser.
4. Follow link to another entity E2. The page shows the references entities of E1 plus the referenced entities of E2.
5. Repeat step 3.
6. Follow linkt to yet another Entity E2: The page shows the references entities of E1, E2 and E3.

Reloading the page (ctrl + r, f5) once shows the correct list of referenced entities.

PR #192, opened 2024-01-31
Drop entity detail cache

TeodoraPavlova:
Related to https://github.com/SUSE/aimaas/issues/191

Maybe you wonder why this is all about? The things I tried before doing the proposed implementation:
- I could've just call deactivated() hook in `ReferencedEntitySelect` and clear the selected. However, then the same problem (described in https://github.com/SUSE/aimaas/issues/191) occurred after selecting another entity, listed as FK in the current detail view.
- Why I didn't exclude the component from keep alive rather than binding key to `<router-view>`? I tried that also, but then the excluded component had no access to the router state.. I also tried to switch `<router-view>` and `<keep-alive>` places, but then we get a warning in the console.
- Why I didn't just did `<router-view :key="$route.fullPath">`? Because it will trigger an instance replacement every time when the route changes.

With the proposed implementation we will be able to easily include/exclude any page which we want or not want to cache. 

I'm open for ideas/suggestions.

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -22,8 +22,8 @@
   </nav>
   <AlertDisplay/>
   <div class="container mt-2">
-    <router-view v-slot="{Component}" :key="routerViewKey">
-      <keep-alive>
+    <router-view v-slot="{Component}">
+      <keep-alive :exclude="nonCachedRouteComponents">
         <component :is="Component"/>
       </keep-alive>
     </router-view>
@@ -60,8 +60,7 @@ export default {
     return {
       activeSchema: null,
       schemaDetails: {},
-      apiInfo: null,
-      routerViewKey: null,
+      apiInfo: null
     }
   },
   provide() {
@@ -84,8 +83,8 @@ export default {
       }
       return _avail_schemas
     },
-    nonCachedRouteNames () {
-      return this.$router.options.routes.filter(route => !route.cached).map(route => route.name);
+    nonCachedRouteComponents () {
+      return this.$router.options.routes.filter(route => !route.cached).map(route => route.component.name);
     },
   },
   methods: {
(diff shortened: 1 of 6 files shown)


jonas-brr:
@TeodoraPavlova I don't understand why the 2nd point doesn't work. Excluding a component from keep-alive shouldn't cause any problems with accessing the router - at least in my experience, maybe something specific to AIMAAS does this (?) For a quick fix I'd really prefer the 2nd bullet point -if it worked - as this solution feels a bit hacky to my taste.

THE proper way imho though would be to do state management the right way - decoupled from the components. A standardized solution could be something like https://pinia.vuejs.org/ - but other ways are ofc possible. Once we'll free the data/state and API interactions from the component hierarchy we'll have no such problems and we will be able to use keep-alive everywhere without issues.

But I admit it'd be a substantially larger time investment.

TeodoraPavlova:
@jonas-brr I totally agree with you, that the solution is a little bit hacky and if we want to do this the 'proper way' (for example using state management library) will take more time..

Regarding why the 2nd point is not working, I'm not sure either.

TeodoraPavlova:
@jonas-brr thanks a lot for confirming my thoughts, I will dive deeper and see what I can do in order to achieve 2nd approach

crazyscientist:
Hey, sorry for being late to the discussion...

This might be a rather fundamental question: Have you considered in your experiments that the `<keep-alive>` in App.vue does not serve an intentional purpose? Maybe it could be removed?

As we have a second `<keep-alive>` in Tabbing.vue: Would it make sense to clear/destroy this cache, when the parent component Tabbing gets unmounted?

TeodoraPavlova:
TBH, I was not aware that we have more than 1 `<keep-alive>` ...

TeodoraPavlova:
> Have you considered in your experiments that the <keep-alive> in App.vue does not serve an intentional purpose? Maybe it could be removed?

Yes, you're right, it can be removed. The removal even resolves the issue. Unpleasantly, I've found that the same problem occurs if you are in entity detail and click on a reference link (the page gets filled with the new entity detail data, but the old references are still there along with the new ones). So, we will still need something like [this](https://github.com/SUSE/aimaas/pull/192/files#diff-bac80d7866040792193851272baf4960472ed704104a62212cb1d42ea472014dR82).

crazyscientist (approved):
>  The removal even resolves the issue.

That is good news :+1: 

> So, we will still need something like [this](https://github.com/SUSE/aimaas/pull/192/files#diff-bac80d7866040792193851272baf4960472ed704104a62212cb1d42ea472014dR82).

I take it, my idea of clearing the `Keep-Alive`'s cache was not viable. And "this" looks simple enough :slightly_smiling_face: :+1:

PR #193, opened 2024-02-02
Update entity list when needed

TeodoraPavlova:
Working on the aimaas frontend in the last days made me think about that we are only doing parent-to-child and child-to-parent data flows (and actually this seems okay in the most of the cases).   

Back in the days before, in another company, we've used a global event bus to transmit events between components. This is now removed in Vue 3 and there is only the ability to transmit events between a parent and a child component, because the global event bus is considered as anti-pattern. I can understand why, I think the abusing it could lead to very unexpected behaviour and bugs which are hard to debug. I see that there are solutions out there like pinia, vuex... but I'm not sure if we want to invest in this at that point.

In one of my previous MRs I've included the ability to refresh the entity list table whenever visible by activated hook. Of course, this solution is not ideal, maybe we would like to refresh it on route change and when something is updated/created. In the last [MR](https://github.com/SUSE/aimaas/pull/192), since we removed the global keep-alive component, the activate hook in entity list is causing unnecessary reloads and can be removed. However, I've noticed that when we switch from entity list to entity create tab, create a new entity and return back to the entity list, the list is not reloaded. This MR introduces a simple solution without 3rd party libraries. 

What do you think about this solution?

### DIFF
--- a/frontend/src/components/EntityList.vue
+++ b/frontend/src/components/EntityList.vue
@@ -51,12 +51,13 @@
 </template>
 
 <script>
-import {computed} from "vue";
+import {computed, watch} from "vue";
 
 import ConfirmButton from "@/components/inputs/ConfirmButton";
 import Pagination from "./layout/Pagination.vue";
 import EntityListTable from "@/components/EntityListTable";
 import SearchPanel from "./SearchPanel.vue";
+import {updateEntityListCount} from "@/store";
 
 export default {
   components: {EntityListTable, Pagination, SearchPanel, ConfirmButton},
@@ -75,8 +76,8 @@ export default {
       default: false
     }
   },
-  activated() {
-    this.getEntities({resetPage: false});
+  created() {
+    watch(updateEntityListCount, (async () => await this.getEntities({resetPage: false})));
   },
   computed: {
     pages: computed(() => {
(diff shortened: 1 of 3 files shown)


jonas-brr:
@TeodoraPavlova nicely explained. I would be strongly against the global prop pass through you mentioned as well (I already forgot how this feature was called :smile: ) - I've done it and it's a sure road to hell IMO.

I would really, really like us to prioritize the work on **a** store solution here (not necessarily pinia). If the app will grow (which I don't know, probably?) these issues will start to pop up more often. Furthermore not extracting this kind of logic into a common place will lead to copy-pasting across different components that share the data and syncing those will be cumbersome and unintuitive (which it probably already is).

I'd say thank you for centralizing at least this one ref, and if it works - good for now :+1: 
...But I'd vote for doing stores :stuck_out_tongue:

TeodoraPavlova:
> I just checked with the current master branch, and the list does get reloaded after creating a new entity.

Yes, this is because of the activated hook presence.

> I would even go so far as to say that we also would like to reload the list of entities in case of other clients posted updates

Yes, I've been thinking about that also, I wasn't sure what are the exact intentions (because of the tabs keep-alive parent), but now I know that we certainly doesn't want to cache this tab.

If I understand you correctly, you're proposing to keep the activated hook and reload whenever the entity list is visible. I agree that this is the most simple solution and with that way we don't need to think when to reload. I'm okay with keeping it. Therefore I will close this MR.

PR #194, opened 2024-02-02
Change activated to mounted where needed

TeodoraPavlova:
Found out that the Pending Reviews page and Create Schema page are not loading (because of the keep-alive removal in [this](https://github.com/SUSE/aimaas/pull/192) MR)


crazyscientist (approved):
Thanks for the rigorous testing and debugging :+1:

Issue #196, opened 2024-02-19
Bulk Add products


celan:
Similar to https://github.com/SUSE/aimaas/issues/90 a Bulk Add functionality is needed.

Use case:
A Product manager would like to add a product, which has several modules with the same life cycle. Putting them in one by one, would be a tedious task.

UI it would be enough to use the standard edit dialog and offer only the required 4 fields for additional products, modules, etc.

PR #201, opened 2024-04-04
Entity Bulk Add (copy Attributes) (#196)

TeodoraPavlova:
resolves https://github.com/SUSE/aimaas/issues/196


frontend/src/components/EntityBulkAdd.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+  <div v-for="num in entityFormComponentsCountRange" :key="`ef-${num}`">
```
jonas-brr, line 2: I'd say having a padding (or flex gap) between the forms would look nicer.

frontend/src/components/inputs/EntityForm.vue
```diff
@@ -6,6 +6,20 @@
...
+        <div class="fw-bold my-2">Values to copy:</div>
+        <div v-for="attr in schema?.attributes || []" :key="`${attr.name}-${attr.id}`" class="mb-1">
+          <label :class="`${attributes[attr.name] ? 'cursor-pointer' : ''}`">
+            {{ attr.name }}
+            <sup v-if="requiredAttrs.includes(attr.name)" class="text-danger me-1"
+                 data-bs-toggle="tooltip" title="This value is required">*</sup>
+            <input type="checkbox" @change="onAttributeCopyChange($event.target.checked, attr.name)"
+                   :checked="editEntity[attr.name]"
+                   :disabled="!attributes[attr.name]"
+                   :class="`${attributes[attr.name] ? 'cursor-pointer' : ''}`" />
```
jonas-brr, line 19: superfluous string format? `:class="attributes[attr.name] ? 'cursor-pointer' : ''"` should do the trick, right?

frontend/src/components/Entity.vue
```diff
@@ -42,6 +43,12 @@ export default {
           icon: "mode_edit",
           tooltip: "Edit/show entity details"
         },
+        {
+          name: "Bulk Add (copy Attributes)",
+          component: EntityBulkAdd,
```
jonas-brr, line 48: This should be wrapped in markRaw or shallowRef (I'm fixing all similar occurences to markRaw in my commit, so I'd prefer markRaw).

frontend/src/components/EntityBulkAdd.vue
```diff
@@ -0,0 +1,57 @@
...
+    <EntityForm :ref="`entity-form-${num}`" :schema="schema"
```
jonas-brr, line 3: I must say I dislike having the "save all" button under each form, it looks strange. Maybe it could be just a single button at the bottom? In addition the UX would benefit from a (smooth) scroll once you click "add more", so that you don't need to scroll manually down each time.

jonas-brr, line 3: I've stumbled upon sth weird while testing this out. If there's something wrong in certain entity forms, "save all" will only create some of them. After fixing what was wrong I now click "save all" again but since the correct items were already created now there are errors again. I'd expect that the created ones would probably disappear from the list leaving only the errored ones?

If there's no error maybe user can be redirected to the entity list (tho there's a risk that the created entities won't be visible on the page)? Maybe just resetting the bulk add form(s) if there's nothing wrong would be enough?
In either case it feels a bit strange nothing changes after a successful save and user stays in the same tab with all the data that was already sent away and successfully saved.

jonas-brr, line 3: I am missing some sort of a remove/cross element to remove a form if I over-click or I just changed my mind and want to remove an entity from the bulk add. Not a big deal, but I've been intuitively looking for it.

frontend/src/components/EntityBulkAdd.vue
```diff
@@ -0,0 +1,57 @@
...
+                :attributes="attributes" :batch-mode="true" @save-all="saveAll"/>
+  </div>
+  <div class="container mt-2">
+    <button class="btn btn-outline-secondary w-100" @click="addEntityForm">
+      <i class='eos-icons'>add_circle</i>
+      Add more
```
jonas-brr, line 9: I noticed that clicking add more always triggers an API request, seems unnecessary since it's all just a copy of the same entity.

frontend/src/components/EntityBulkAdd.vue
```diff
@@ -0,0 +1,57 @@
...
+    </button>
+  </div>
+</template>
+
+<script>
```
jonas-brr, line 14: could have been a `<script setup>` :cry: :stuck_out_tongue:

frontend/src/components/EntityBulkAdd.vue
```diff
@@ -0,0 +1,57 @@
...
+      required: true
+    },
+    attributes: {
+      type: Object,
+      required: true,
+    }
+  },
+  data() {
+    return {
+      entityFormComponentCount: 1,
```
jonas-brr, line 32: This is not very nice, it's a shame that entity form isn't implemented as a v-model. It'd be very easy to hold a list of entities inside this components and `<EntityForm v-for="entity in entities" v-model="entity" />` and having a single "save all" button that could easily manage the bulk save from here. I'd say this would be the way to go, solving most of my other suggestions, but it's a probably a lot of work to refactor :disappointed:

frontend/src/components/EntityBulkAdd.vue
TeodoraPavlova: Agree for both, but unfortunately for the save all button additional refactoring should be made.. the entity form is used for bulk edit also and the behavior is the same there

frontend/src/components/EntityBulkAdd.vue
TeodoraPavlova: I agree that this would solve most of the suggestions.. I guess we can do the refactoring while switching to composition API? Or even sooner? IDK, I will think about it

jonas-brr: I guess there's no need to rewrite the entire component to composition API, only implementing the v-model in it, and fix the component usages throughout the repo (shouldn't be that many). I wouldn't say it's not a huge refactoring deserving its own issue/effort and I'd personally be fine updating this one while working on this issue (since it'd enable me to implement it properly much faster). In addition the issue is a 7, so it should be ok to spend some more time on it. But I can't speak for the others and it's ultimately up to you :slightly_smiling_face:

frontend/src/components/EntityBulkAdd.vue
jonas-brr: yeah... that happens if forms aren't done using v-model and handle the actions (with buttons) internally :disappointed:
I'd say if you decide for not doing any refactoring, leave the buttons as is rather than doing some crazy if/else tweaks in the form component.

frontend/src/components/EntityBulkAdd.vue
TeodoraPavlova: yeah, I believe that's because the nested components do requests.. here additional refactoring should be made also

frontend/src/components/EntityBulkAdd.vue
TeodoraPavlova: Nice suggestion ;) We should do this across the app, since I didn't found any occurrences.. do you think that this deserves to be a new issue?

jonas-brr: I don't think so.. I wouldn't like to dedicate extra effort to refactor components just because I like the style more. I still think it is going to make the code better - so (if it was up to me :stuck_out_tongue: ) I'd like us to slowly transition - i.e. if someone writes a new component from scratch: composition api (hence the comment), or if someone does some extended component refactoring that might be simplified with composition api: go for it. But let's not rewrite components 'just because' - it might not be the best usage of the team's time.

jonas-brr: ...and don't feel pressured, I understand that it's something new that requires a bit extra effort to ease into, so this was just a tiny, friendly nudge :slightly_smiling_face:

frontend/src/components/EntityBulkAdd.vue
jonas-brr: yes, it's no cause by your changes directly, tho I still felt like mentioning it since it directly impacts the new feature

frontend/src/components/EntityBulkAdd.vue
TeodoraPavlova: Ah, I see what you mean now.. You got a point. I will try to do it, It will be nice exercise and reasonable investment :)

frontend/src/components/EntityBulkAdd.vue
TeodoraPavlova: Yeah, fair point.. I would like to combine it with future refactoring of entity form in another branch and issue (if you don't mind). I will leave myself a note in order to not forget to create the task

frontend/src/components/EntityBulkAdd.vue
```diff
@@ -29,24 +36,45 @@ export default {
...
-    addEntityForm() {
-      ++this.entityFormComponentCount;
+    addNewItem() {
+      this.addEntityForm().then(() => {
+        document.getElementById(`form-${this.entityForms.length - 1}`).scrollIntoView();
+      });
     },
+    generateEntityForm() {
+      return {
+        component: markRaw(EntityForm),
```
jonas-brr, line 71: cool workaround, I'd say that's much better :+1:

frontend/src/components/EntityBulkAdd.vue
```diff
@@ -0,0 +1,92 @@
...
+      Add more
+    </button>
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'EntityBulkAdd'
+}
+</script>
```
crazyscientist, line 26: I'm a bit new to the composition API, but if there is a `<script setup>` tag, this `<script>` tag becomes obsolete, right?

frontend/src/components/EntityBulkAdd.vue
```diff
@@ -0,0 +1,92 @@
...
+    const component = x[1];
+
+    if (refName && component) {
+      const resp = await component.createEntity();
+      if (resp?.id) {
+        successIds.push(refName);
+        component.editEntity.name = null;
+        component.editEntity.slug = null;
+      }
+    }
```
crazyscientist, line 71: Should we handle the case of this condition being false? E.g. via a warning in the JS console? Or I'm I too cautious here?

frontend/src/components/Entity.vue
```diff
@@ -91,12 +92,14 @@ export default {
         this.entity = null;
       }
     },
-    async onUpdate() {
-      await this.updateEntity();
+    async onUpdate(entity) {
+      if (entity) {
+        this.entity = entity;
+      }
```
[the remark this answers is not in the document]
TeodoraPavlova: If I'm not missing something, If there is no entity emitted, the EntityForm was used for creation of a new entity and there is no need to update

frontend/src/components/EntityBulkAdd.vue
TeodoraPavlova: > I'm a bit new to the composition API
Yeah, me too and it turned out that we need that, because otherwise the `EntityBulkAdd` was not rendered. 

I found [this thread](https://stackoverflow.com/questions/67669820/how-to-define-component-name-in-vue3-setup-tag) which helped me resolving it..

The strange thing here is that I saw a comment in the thread:

> Update from vue site "Since version 3.2.34, a single-file component using <script setup> will automatically infer its name option based on the filename, removing the need to manually declare the name  

I see that we're using version `3.3.6`, but if we not specify name in separate script tag, it does not work. It could be related with the fact that the rest of the app still uses options API, but IDK..

frontend/src/components/EntityBulkAdd.vue
TeodoraPavlova: I think we don't need it. I'm collecting only components which have successful response in order to exclude them later. The ones, which caused errors will still remain displayed and a message will pop up to the user.

frontend/src/components/EntityBulkAdd.vue
jonas-brr: I've stumbled upon the same thing - i.e. the name attribute being removed - but it shouldn't matter if the rest is still options or not, the name should be inferred from the file name (as you mention it) and available as if you'd written it explicitly.

jonas-brr: isn't the culprint a wrong import? Shouldn't `import EntityBulkAdd from "@/components/EntityBulkAdd.vue";` be without the `.vue` at the end? :thinking:

TeodoraPavlova: Yes, that's it! Thank you!

TeodoraPavlova: I should blame the auto import feature from JetBrains for this :laughing:

frontend/src/components/inputs/EntityForm.vue
```diff
@@ -110,6 +124,14 @@ export default {
     batchMode: {
       type: Boolean,
       default: false
+    },
+    attributes: {
+      type: Object,
+      required: false,
+    },
+    id: {
```
jonas-brr, line 132: This is something that only an external component needs. imho it shouldn't be here, but if that's the case maybe at least a comment with where/why it is needed could be helpful? This will be really hard to understand later (especially without TS).

frontend/src/components/EntityBulkAdd.vue
```diff
@@ -0,0 +1,92 @@
...
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'EntityBulkAdd'
+}
+</script>
+
+<script setup>
```
jonas-brr, line 28: :+1:

jonas-brr:
@TeodoraPavlova @crazyscientist  I hate imposing yet another thing, but would you be against using prettier to format the source code? I wouldn't be against if someone'd like to tweak the default settings, but if not we could all just enable the prettier plugin to reformat on save and call it a day (?) :thinking:

frontend/src/components/EntityBulkAdd.vue
crazyscientist: Nicely spotted. I had completely overlooked the `.vue` suffix.

crazyscientist:
> would you be against using prettier to format the source code?

@jonas-brr Not at all. So far, PyCharm (without plugin) seemed to do a well enough job for me. Would you like to report this as a new issue?

frontend/src/components/EntityBulkAdd.vue
```diff
@@ -0,0 +1,92 @@
+<template>
+  <div v-for="(form, index) in entityForms" :key="form.props.id">
+    <div :id="`form-${index}`">
+      <div :class="`${entityForms.length < 2 && 'd-none'} ${index > 0 && 'mt-5 border-top border-light'} row`">
+        <div class="col">
+          <button type="button" :class="`btn-close float-end ${index > 0 ? 'my-3': 'mb-3'}`"
+                  @click="closeForm(form.props.id)"/>
+        </div>
+      </div>
+      <component :is="form.component" @save-all="saveAll" v-bind="form.props"
```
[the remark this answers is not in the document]
TeodoraPavlova: 0. I already created a refactoring task
1. Ok,  makes sense
2. I already tried several things (which were without an 'id' and not working as expected) and I don't remember specifics anymore, but your suggestion is accurate and I will try to get rid of the id.

frontend/src/components/Entity.vue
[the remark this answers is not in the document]
TeodoraPavlova: https://github.com/SUSE/aimaas/pull/201/files#diff-cdc34d04ddeed35251b6f776db2bd7a27384f0b8889803cda15f598a419bc2a4R267

If you want we can emit an update event if only in edit mode? I was kind of afraid doing this specifically, since I could break something which use it even in create mode

jonas-brr:
> PyCharm (without plugin) seemed to do a well enough job for me

@crazyscientist after trying it out ( control + alt + L (?) ), pyCharm doesn't seem to be clever enough to understand .vue (without the vuejs plugin - not available in community edition). However even with pure html/js files, it looks like it doesn't a good job there either.

I'd say prettier would be a good thing for having consistent formatting (though I might be the only one bothered by that :smile: ).

Yes, if you (and others probably) aren't using VS code, I could create a tiny issue for adding prettier to npm scripts for starters, and maybe adding a pipeline check(s) later (?)

frontend/src/components/EntityBulkAdd.vue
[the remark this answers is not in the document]
jonas-brr: Yeah, I know, you mentioned it, it just feels a bit like doing things twice. But thanks anyway for filing the following issue :slightly_smiling_face:

ad2: I thought so, if you already spent some time on it, don't worry about it. Just a tiny hint would be enough.

crazyscientist:
> I could create a tiny issue for adding prettier to npm scripts for starters, and maybe adding a pipeline check(s) later (?)

@jonas-brr That sounds like a very good plan :+1:

frontend/src/components/Entity.vue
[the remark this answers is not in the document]
crazyscientist: There are also the deletion and restoration of deleted entities, which would provide a falsy `entity` argument. So, I would be inclined to follow your original approach :+1:

TeodoraPavlova:
@jonas-brr As you suggested, I'm not bothered by that and PyCharm does the job for me. However, there could be annoying consequences of formatters like messing up the git history (everything looks like it's done by 1 person or at least the last changes and you have to dig). The other thing is merge conflicts..

But I'm not against formatters, I just have bad memories :laughing: If this is important for you, please go ahead, I think in this case the things I've suffered from before are not relevant (not a lot of pending MRs and the git history does not matter, especially if we're planning to migrate to composition API)

TeodoraPavlova:
@crazyscientist @jonas-brr is still anything expected from me here?

frontend/src/components/inputs/EntityForm.vue
```diff
@@ -110,6 +129,10 @@ export default {
     batchMode: {
       type: Boolean,
       default: false
+    },
+    attributes: {
```
crazyscientist, line 133: If I read teh code in EntityBulkAdd.vue correctly, an entity data object is passed in as `attributes` property. Would it make sense to reuse the `entity` property instead?

frontend/src/components/Entity.vue
```diff
@@ -64,13 +71,14 @@ export default {
     bindArgs() {
       return [
         { schema: this.activeSchema, entity: this.entity },
+        { schema: this.activeSchema, attributes: this.entity },
```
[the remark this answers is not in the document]
TeodoraPavlova: The entity is used for editing and it does different requests, that's why I created an additional property for attributes. If there is no entity, the EntityForm automatically sets the editEntity attributes to null. It's probably a little bit confusing here, since I'm using the whole entity.. 

I can try to refactor a little bit, so the EntityForm can send an update request only if there is entity id and remove the attributes property. If I go this direction we will still need another property for conditional displaying of the attribute checkboxes. But I guess this will be more readable and it will make more sense.

frontend/src/components/inputs/EntityForm.vue
TeodoraPavlova: Yes, same as above -> https://github.com/SUSE/aimaas/pull/201#discussion_r1574443807

frontend/src/components/EntityBulkAdd.vue
```diff
@@ -0,0 +1,98 @@
...
+      attributes: props.attributes,
+      schema: props.schema,
+      batchMode: true,
+    }
+  }
+}
+
+async function saveAll() {
+  let successfullySaved = [];
+  const promises = Object.entries(entityFormRefs.value).map(async ref => {
```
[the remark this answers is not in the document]
TeodoraPavlova: Nice catch, thanks!

frontend/src/components/Entity.vue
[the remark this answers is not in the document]
crazyscientist: I think I need another look at the code without getting interrupted.

If you think this will increase readability and if you are willing to refactor like that, it will be much appreciated :slightly_smiling_face:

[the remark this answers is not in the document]
TeodoraPavlova: > The entity is used for editing and it does different requests, that's why I created an additional property for attributes.

Hey, sorry, I want correct myself a bit - today, when I went through the code, I realized that the above explanation which I gave is not relevant since I use directly createEntity() function from EntityForm component..

crazyscientist (approved):
LGTM

PR #202, opened 2024-04-05
Fix DateTime attribute field value does not get rendered on Entity edit

TeodoraPavlova:
When there is a value returned from the server for datetime attribute, on Entity detail page the DateTime component does not render it due to wrong format.

### DIFF
--- a/frontend/src/components/inputs/DateTime.vue
+++ b/frontend/src/components/inputs/DateTime.vue
@@ -2,7 +2,7 @@
   <BaseInput :label="label" :args="args" :vertical="vertical" :required="required">
     <template v-slot:field>
       <input class="form-control" type="datetime-local"
-             :value="modelValue && convertToDateTimeLocalString(modelValue)"
+             :value="value"
              @input="onInput"
              v-bind="args"/>
     </template>
@@ -16,6 +16,11 @@ export default {
   name: "DateTime",
   components: {BaseInput},
   props: ["label", "modelValue", "args", "vertical", "required"],
+  computed: {
+    value() {
+      return this.modelValue && this.convertToDateTimeLocalString(this.modelValue);
+    }
+  },
   methods: {
     onInput(event) {
       this.$emit("update:modelValue", event.target.value);
--- a/frontend/src/components/inputs/DateTime.vue
+++ b/frontend/src/components/inputs/DateTime.vue
@@ -1,7 +1,9 @@
 <template>
   <BaseInput :label="label" :args="args" :vertical="vertical" :required="required">
     <template v-slot:field>
-      <input class="form-control" type="datetime-local" :value="modelValue" @input="onInput"
+      <input class="form-control" type="datetime-local"
+             :value="modelValue && convertToDateTimeLocalString(modelValue)"
+             @input="onInput"
              v-bind="args"/>
     </template>
   </BaseInput>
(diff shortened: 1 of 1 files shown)


frontend/src/components/inputs/DateTime.vue
```diff
@@ -1,7 +1,9 @@
 <template>
   <BaseInput :label="label" :args="args" :vertical="vertical" :required="required">
     <template v-slot:field>
-      <input class="form-control" type="datetime-local" :value="modelValue" @input="onInput"
+      <input class="form-control" type="datetime-local"
+             :value="modelValue && convertToDateTimeLocalString(modelValue)"
```
jonas-brr, line 5: This feels wrong for. I think we shouldn't modify the value in here like this, but rather take care of the modelValue in the script section - it'll be more intuitive/readable. The main problem I have with this component is that I don't know what the model value type is supposed to be. Is  it `Date`? String?

In either case I'd rather force the model value to always be Date and than handle the **string -> Date** conversion in the `onInput`, which will self-correct the model value to be again of type Date... simply:

```
onInput(event) {
  const resolvedDate = new Date(event.target.value);
  this.$emit('update:modelValue', resolvedDate);
}
```

...because datetime input can handle the **Date -> string** conversion internally on its own, right?

P.S. I hope that

TeodoraPavlova: Everything works correctly on input. The problem occurs when there is a value coming from the backend and it's formatted like "2024-04-08T15:15:00+03:00" which the datetime-local input does not like and needs to be reformatted in order to get displayed.

> The main problem I have with this component is that I don't know what the model value type is supposed to be. Is it Date? String?

String

> ...because datetime input can handle the Date -> string conversion internally on its own, right?

Actually, it only accepts string as value :/

> This feels wrong. I think we shouldn't modify the value in here like this, but rather take care of the modelValue in the script section

Yes, you got a point, I agree.

jonas-brr: huh.. I tried this out and it worked fine.. the thing is that when you pass the incoming string to `new Date()` it parses without any issues and then you can safely pass the `Date` type to the input directly.

jonas-brr: weird, I can't recall what I did... I must have done something wrong. My apologies, it looks like the input can't indeed handle the Date type. So we can't make this any nicer.

jonas-brr (approved):
lgtm

frontend/src/components/inputs/DateTime.vue
```diff
@@ -14,10 +16,28 @@ export default {
...
+        date = new Date(date)
+      }
+
+      const year = date.getFullYear();
+      const month = (date.getMonth() + 1).toString().padStart(2, "0");
+      const day = date.getDate().toString().padStart(2, "0");
+      const hours = date.getHours().toString().padStart(2, "0");
+      const minutes = date.getMinutes().toString().padStart(2, "0");
+
+      return `${year}-${month}-${day}T${hours}:${minutes}:00`;
```
crazyscientist, line 39: If the [docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local#value) are to be trusted, the expected format for the return value is `YYYY-MM-DDThh:mm`. So we could simplify to:

```suggestion
      return date.toISOString().substring(0, 16);
```

Assuming that you don't find this too opaque.

TeodoraPavlova: I like it, it's even more readable. Thanks!

crazyscientist (approved):

PR #204, opened 2024-04-16
Rearranging UI elements

crazyscientist:


crazyscientist:
The result UI:

grafik

frontend/src/components/inputs/EntityForm.vue
```diff
@@ -5,19 +5,24 @@
...
-            {{ attr.name }}
-            <sup v-if="requiredAttrs.includes(attr.name)" class="text-danger me-1"
-                 data-bs-toggle="tooltip" title="This value is required">*</sup>
-            <input type="checkbox" @change="onAttributeCopyChange($event.target.checked, attr.name)"
-                   :checked="editEntity[attr.name]"
-                   :disabled="!attributes[attr.name]"
-                   :class="attributes[attr.name] ? 'cursor-pointer' : ''" />
-          </label>
+      <div class="d-flex justify-content-between align-items-end">
+        <h3 class="mt-3">Attributes</h3>
```
TeodoraPavlova, line 9: It looks better than mine, thanks!

I will propose a small suggestion - if the attributes are too many, it looks like:
image

I don't insist on this one, but I think it will look better if we do:

```suggestion
        <h3 class="align-self-start">Attributes</h3>
```

crazyscientist: Yes. But also like the spacing above the "Attributes" heading. But that can be moved somewhere else. Let me push another commit...

PR #205, opened 2024-04-17
Refresh entity list on delete/restore

TeodoraPavlova:
How to reproduce - go to entity list page and delete/restore entity. The list does not get refreshed (without manual page reload).

### DIFF
--- a/frontend/src/components/EntityList.vue
+++ b/frontend/src/components/EntityList.vue
@@ -155,8 +155,8 @@ export default {
       this.selected = this.$refs.selector.getSelected();
     },
     onDeletion() {
-      const promises = this.selected.map(eId => {
-        this.$api.deleteEntity({
+      const promises = this.selected.map(async eId => {
+        await this.$api.deleteEntity({
           schemaSlug: this.schema.slug,
           entityIdOrSlug: eId
         });
@@ -164,8 +164,8 @@ export default {
       Promise.all(promises).then(() => this.getEntities({resetPage: true}));
     },
     onRestoration() {
-      const promises = this.selected.map(eId => {
-        this.$api.restoreEntity({
+      const promises = this.selected.map(async eId => {
+        await this.$api.restoreEntity({
           schemaSlug: this.schema.slug,
           entityIdOrSlug: eId
         });


crazyscientist:
Hi there, I just checked the reproduction steps and come to a little different conclusion: 

So, I selected 4 entities for deletion and the app sent 4 DELETE requests and 1 GET request in order to update the list:

grafik

In my particular case, the updated list kept showing one of the deleted entities. So, I would conclude, that there is a race condition and that the app is not waiting for the completion of all DELETE requests prior to the GET request.

frontend/src/components/EntityList.vue
```diff
@@ -155,17 +155,17 @@ export default {
       this.selected = this.$refs.selector.getSelected();
     },
     onDeletion() {
-      const promises = this.selected.map(eId => {
-        this.$api.deleteEntity({
+      const promises = this.selected.map(async eId => {
+        await this.$api.deleteEntity({
```
crazyscientist, line 159: So, the original idea was to trigger all the async. API calls to delete entities and store their promises in the variable.

Later (in line 164) they are all awaited and only afterwards the API function to get entities is called.

To me it looks like the `.map()` call does not return promises, which makes the call to `Promise.all()` behave in an unintended way.

Maybe this would do the trick:

```suggestion
      const promises = this.selected.map(eId => {
        return this.$api.deleteEntity({
```

frontend/src/components/EntityList.vue
```diff
@@ -155,17 +155,17 @@ export default {
...
           entityIdOrSlug: eId
         });
       });
       Promise.all(promises).then(() => this.getEntities({resetPage: true}));
     },
     onRestoration() {
-      const promises = this.selected.map(eId => {
-        this.$api.restoreEntity({
+      const promises = this.selected.map(async eId => {
+        await this.$api.restoreEntity({
```
crazyscientist, line 168: Like above, I think this would restore the intended behavior

```suggestion
      const promises = this.selected.map(eId => {
        return this.$api.restoreEntity({
```

frontend/src/components/EntityList.vue
TeodoraPavlova: Yes, you're right! Thanks!

PR #207, opened 2024-04-30
Load schema on demand (fixes #203)

### LINKED ISSUE #203: Schema attribute changes not reflected in Entity detail page
If you change the schema attributes and after that go to entity detail page, you won't see schema changes reflected. The changes are visible after page reload.

I thought that we can update the `activeSchema` value with triggering an update in `SchemaEdit` component (if we provide a function, which will do the update in `<App/>`, inject it into `<SchemaEdit/>` and invoke it). It turned out that the `activeSchema` object gets updated, but provide() does not get called again after updating the object. 

It could be possible that at this point this can not be resolved properly (without javascript page reload) and it could be issue related with the usage of the options API. In theory, it should be reactive and the provide() should get called after.

If there is no other easy solution at this point, a page reload (on successful response) will do the job.

TeodoraPavlova:
resolves https://github.com/SUSE/aimaas/issues/203


frontend/src/components/Entity.vue
```diff
@@ -1,113 +1,115 @@
...
+  } else {
+    entity.value = null;
+  }
+}
+
+async function onUpdate(entity) {
+  if (entity) {
+    entity.value = entity;
+  }
+}
```
crazyscientist, line 97: I have not checked whether this function behaves as intended, but from the looks I would say no:

It looks like this function is setting the `value` attribute of the object in the argument to the object in the argument.

Maybe you would like to rename the argument of this function?

frontend/src/components/Schema.vue
```diff
@@ -1,92 +1,104 @@
...
+});
+
+onUpdated(() => {
+  if (activeSchema.value?.name) {
+    document.title = activeSchema.value.name;
+  }
+});
+
+async function onSchemaUpdate() {
+  await getSchema();
```
crazyscientist, line 90: Similar to the watcher below, you might want to consider setting `loading.value` to `true` and `false` before and after the call to `getSchema` respectively :thinking:

crazyscientist (changes requested):

frontend/src/components/Entity.vue
TeodoraPavlova: Nice catch, thanks!

frontend/src/components/Schema.vue
TeodoraPavlova: Yes, thank you!

frontend/src/components/SchemaEdit.vue
```diff
@@ -49,6 +49,7 @@ export default {
   name: "SchemaEdit",
   components: {EditAttributes, TextInput, Checkbox},
   props: ["schema"],
+  emits: ["update-schema"],
```
crazyscientist, line 52: Do we use this event somewhere?

TeodoraPavlova: Yup https://github.com/SUSE/aimaas/pull/207/files#diff-04b105b5ae587de59f4bab07a7c490d96d52b13ef8785ef5f138172a22eb5b95R16

crazyscientist: Ah :bulb: My bad, I did not check for the CamelCase version. Would it make to use the same formatting here as in the other file?

TeodoraPavlova: yes, it works and it will be less confusing :+1:

crazyscientist: Cool, thank you :slightly_smiling_face:

PR #215, opened 2024-08-26
fix: cascade delete permissions (closes #200)

### LINKED ISSUE #200: Deleting group doesn't cascade delete permissions
As discovered during [#197](https://github.com/SUSE/aimaas/pull/198/files/28a588d8c5b28ebf1f6a93d0fa6869a42fa06b16#diff-72c3c3cf76d0c9bdc789549b4e0f832a9109171d818ccb0e942f26815c70c0d3) - deleting a group doesn't cascade delete related permissions. This is probably true for other similar models/relationships as well.

jonas-brr:


jonas-brr:
closes #200

backend/tests/test_crud_auth.py
```diff
@@ -320,3 +319,29 @@ def test_revoke_permission__not_existent(self, dbsession: Session):
...
+        dbsession.commit()
+
+        permissions_count = (
+            dbsession.query(Permission)
+            .filter_by(recipient_type=recipient_type, recipient_id=locals().get(attr_name).id)
+            .count()
+        )
+        assert permissions_count == 0
+
+        x = 1
```
crazyscientist, line 347: This looks like it can be removed...

```suggestion
```

jonas-brr: :man_facepalming: oops, awkward break-point left behind, thanks

PR #221, opened 2025-03-10
fix: add trim modifier to text input (closes #220)

### LINKED ISSUE #220: Strip leading/trailing empty spaces
### Problem
Not stripped input values trigger a very cryptic error:

```
body,attributes,0,name: Attribute name must be a valid Python identifier and must not start/end with underscore
```

### Expectation
Trailing/leading spaces left in an input by accident should be stripped during the data validation.

jonas-brr:


frontend/src/components/inputs/TextInput.vue
```diff
@@ -1,35 +1,44 @@
...
-    onInput(event) {
-      this.$emit("update:modelValue", event.target.value);
-    },
-  },
-};
+const modelValue = defineModel({ required: true });
+
+const props = defineProps([
+  "label",
+  "string",
```
TeodoraPavlova, line 34: Maybe a stupid question, but what is "string" used for?

backend/tests/conftest.py
```diff
@@ -350,7 +350,7 @@ def delete(self, url, *, params=None, headers=None, cookies=None, auth=USE_CLIEN
             headers=headers,
             cookies=cookies,
             auth=auth,
-            follow_redirects=self._choose_redirect_arg(follow_redirects, allow_redirects),
+            follow_redirects=False,
```
TeodoraPavlova, line 353: Sorry, maybe I need a bit of a background here, why this is changed?

.github/workflows/default.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         with:
           python-version: "3.10"
       - name: Cache pip
-        uses: actions/cache@v2
+        uses: actions/cache@v4
```
TeodoraPavlova, line 25: I won't mind updating these in this MR, but generally, I would prefer doing this in a separate one.

jonas-brr: yeah.. I just wanted to see if that solves it, and it did.. Since it's a one liner I kept it.

backend/tests/conftest.py
jonas-brr: That's some redirect logic magic for the custom test client (only there so we can send deletes with `json`). As the client changes this custom one gets broken. This time it is the `_choose_redirect_arg` method - it got removed from the base starlette test client.

Since we don't care about redirects in our tests, I just removed it.

frontend/src/components/inputs/TextInput.vue
jonas-brr: tbf, I have no idea :smile: I just rewrote it to composition API as-is.

jonas-brr: but yes, bad name for a prop imho...

.github/workflows/default.yml
jonas-brr: https://github.com/SUSE/aimaas/pull/223

frontend/src/components/inputs/TextInput.vue
TeodoraPavlova: but the string prop is added by you? :laughing:

backend/tests/conftest.py
TeodoraPavlova: Ok, thanks for the info

.github/workflows/default.yml
TeodoraPavlova: ok, thanks

frontend/src/components/inputs/TextInput.vue
jonas-brr: really?? Oh.. it must be some garbage from the types I added on autopilot and then realized the codebase is not typed..

jonas-brr: damn.. well spotted, thanks!

TeodoraPavlova: no worries :beers:

TeodoraPavlova (approved):

